PR #4, opened 2017-06-15
fix typo and extra connect=False param

gilbertozp:


gilbertozp:
already fixed, hadn't fetched from upstream

PR #7, opened 2017-07-12
Do not wait until all items are processed to update targets

dwinston:
@matk86 please look over and ensure MPI operation is similar before merge.


matk86:
done

PR #12, opened 2018-01-23
adds grouping function and test to make aggregation-based builds

montoyjh:
Adds a grouping function to mongolike stores.  Mongolike.groupby(keys, properties, criteria) constructs and executes a simple aggregation pipeline where you filter, project, and group by a set of keys.


dwinston:
LGTM

PR #13, opened 2018-01-31
Issue 8 vault store

shreddd:


shreddd:
Stupid codacy doesn't like unused classes. But isn't that the point of a library??

shyamd:
It's an upstream problem in pyLint that was fixed ~ Dec 2017, but has not propagated through to Codacy. Let me know when you're ready to commit.

PR #14, opened 2018-01-31
PR for generic Schema class

mkhorton:
This is a very generic `Schema` and a more useful subclass `StandardSchema`.

A `Schema` can be attached a store, and its `is_valid(doc)` method will be called when a doc is attempted to be inserted.

The *only* method a `Schema` has to implement is `is_valid(doc)`.

A `Schema` also has a `.strict` property to tell the `Store` how it wants to be used: if strict, the Store will raise a `ValueError` if validation fails, if not strict, build will just continue and an error logged. In either case, only valid documents will be inserted to the `Store`.

The `StandardSchema` has some useful defaults to check against a provided JSON schema definition (this is a standard, lots of tools exist to help make one, and--more importantly--if you have a valid JSON schema it can be attached to a Mongo collection as of Mongo 3.6+ so it allows really fast document level validation inside Mongo itself too), and `StandardSchema` also checks that certain keypaths are MSONable back to Python objects.

As an example in the Topology builder, I create a `TopologySchema(StandardSchema)` that looks like this:

```
class TopologySchema(StandardSchema):

    @property
    def schema(self):
        return {
            "type": "object",
            "properties":
                {
                    "task_id": {"type": "string"},
                    "method": {"type": "string"},
                    "status": {"type": "boolean"}
                },
            "required": ["task_id", "method", "status"]
        }

    @property
    def msonable_keypaths(self):
        return {"graph": StructureGraph}
```


mkhorton:
Ok, ready :)

PR #15, opened 2018-02-07
Rename Schema to Validator

mkhorton:
Reasons:

1) 'Validator' is a more accurate description of what it does.
2) Schemas sound scary.
3) You can write a valid validator without writing a schema (though a formal schema is encouraged for the StandardValidator).

### DIFF
--- a/maggma/stores.py
+++ b/maggma/stores.py
@@ -32,7 +32,7 @@ class Store(MSONable, metaclass=ABCMeta):
         self.lu_field = lu_field
         self.lu_type = lu_type
         self.lu_func = LU_KEY_ISOFORMAT if lu_type == "isoformat" else (identity, identity)
-        self.schema = None
+        self.validator = None
 
     @property
     @abstractmethod
--- a/maggma/stores.py
+++ b/maggma/stores.py
@@ -206,6 +206,9 @@ class Mongolike(object):
 
         bulk = self.collection.initialize_ordered_bulk_op()
 
+        # filter out None values
+        docs = [doc for doc in docs if doc]
+
         for d in docs:
 
             d = jsanitize(d, allow_bson=True)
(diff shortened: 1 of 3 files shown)


maggma/stores.py
```diff
@@ -206,6 +206,9 @@ def update(self, docs, update_lu=True, key=None):
 
         bulk = self.collection.initialize_ordered_bulk_op()
 
+        # filter out None values
+        docs = [doc for doc in docs if doc]
```
dwinston: @mkhorton I think it'd be better to do this in a `Runner`'s `processor`. There's a `filter(None, chunk)` of processed items before handing off to `update_targets` in [`maggma.runner.SerialProcessor`](https://github.com/materialsproject/maggma/blob/68e09b65ba6852854428e4ada2380413b6270b51/maggma/runner.py#L63), but it looks like similar filtering got left out in @shyamd's refactor of [`maggma.runner.MultiprocProcessor`](https://github.com/materialsproject/maggma/blob/68e09b65ba6852854428e4ada2380413b6270b51/maggma/runner.py#L203) to use `Pool.imap`. I think it'd be best to filter for `None`s there, e.g. passing `list(filter(None, items))` to the `update_targets` call there instead of `items`.

Also, it's okay now, but in future it'd be cleaner to have this as a separate small PR because it's a separate issue than renaming Schema to Validator.

Thanks for discovering this!

mkhorton: OK, sounds good! Makes sense this had already been addressed elsewhere. Will revert this commit.

(And yes, agreed, separate PRs definitely preferred 👍 )

shyamd:
Can you fix the codacy issues? Ignore the unused variable __class__  pylint error, i just turned that rule off.

shyamd:
Thanks!

PR #16, opened 2018-02-08
groupby change for memory/jsonstore

montoyjh:
mongomock doesn't appear to cleanly support aggregation, so this moves groupby into MongoStore, rather than Mongolike, and adds a notImplementedError into the groupby functions on MemoryStore (and JSONStore).


shyamd:
Can you fix the code quality review?

montoyjh:
Should be good.

shyamd:
Thanks

PR #17, opened 2018-02-09
New Multiprocessor and MPI Processor

shyamd:
## Overhauled the Multiprocessor and MPI Processors.

Multiprocessor had lots of little issues that were fixed with the pool imap ,but new inefficiencies cropped up in their place. The current paradigm is based off using concurrent.futures process pool, some threading, and data synchronization to be able to pull, process, and push data as efficiently as resources will allow. 

The previous MPI processor just didn't work. New method uses a lot of the same paradigms as Multiprocessor, with a bit of added complexity for the MPI capabilities. For now logging seems to pass through OpenMPI 3.0 fine. If there is an issue found in the future, we might need a separate logging hook that explicitly passes it via the MPI communicator.


dwinston:
This should have tests. Looking at `mpi4py`'s [.travis.yml](https://github.com/mpi4py/mpi4py/blob/master/.travis.yml), they have a simple install script `./conf/ci/travis-ci/install-mpi.sh` in their repo that we can use so that tests run on Travis.

We need confidence going forward (post-merge) that any changes to the MPI code don't break things.

PR #20, opened 2018-04-13
GridFS Store update: use metadata field, update removes old file(s)

dwinston:
* Change GridFS store to use "metadata" field as per official spec
* Remove old files matching GridFS store key on update

### DIFF
--- a/maggma/stores.py
+++ b/maggma/stores.py
@@ -5,6 +5,7 @@ Stores are a default access pattern to data and provide
 various utillities
 """
 from abc import ABCMeta, abstractmethod
+import copy
 from datetime import datetime
 import json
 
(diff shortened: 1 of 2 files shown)


maggma/stores.py
```diff
@@ -561,13 +582,16 @@ def query(self, properties=None, criteria=None, **kwargs):
...
             data = f.read()
+            metadata = f.metadata
             try:
-                json_dict = json.loads(data)
-                yield json_dict
+                data = json.loads(data)
             except:
-                yield data
+                pass
+            yield dict(data=data, metadata=metadata)
```
shyamd: It shouldn't actually be necessary to yield back the metadata, since the keys put into the gridfs  metadata will either already be in the doc or be an accessory for GridFSStore such as 'compression'

maggma/stores.py
```diff
@@ -679,8 +693,17 @@ def update(self, docs, update_lu=True, key=None):
...
+                d[self.lu_field] = datetime.utcnow()
+
+            data = json.dumps(jsanitize(d["data"])).encode("UTF-8")
+            metadata = {self.lu_field: d[self.lu_field]}
+            metadata.update(search_doc)
+            self.collection.put(data, metadata=metadata)
+            self.transform_criteria(search_doc)
+            for fdoc in (self._files_collection.find(search_doc, ["_id"])
+                    .sort("uploadDate", -1).skip(1)):
+                self.collection.delete(fdoc["_id"])
```
shyamd: I think the alternative stores should still 'appear' to operate as MongoStores do. It shouldn't be necessary to know this is a GridFS store or writing builders will become obnoxious. Rather, this store should be given a set of keys, just like normal update that it will store in the metadata for searching: Maybe this is part of it's ensure_indexes ? 

GridFS has also been historically used with compression. This should be stored in the metadata as well. We tend to use "compression" : "zlib" and not much else.

PR #23, opened 2018-04-27
Jointstore

montoyjh:
A store that can use the lookup aggregation functionality to query sub collections along with a master collection such that documents in sub collections with matching keys as the master will be treated as subdocuments of master docs.

e. g. you have two collections, master and sub1:

```
js = JointStore(...)
docs = query([key1, sub1.key2], {"sub1.key2": {"$gt": 3}})
```
etc.

### DIFF
--- a/maggma/advanced_stores.py
+++ b/maggma/advanced_stores.py
@@ -22,7 +22,7 @@ try:
     import boto3
     import botocore
     boto_import = True
-except:
+except ImportError:
     boto_import = False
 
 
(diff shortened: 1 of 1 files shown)


dwinston:
(Probably) beyond the scope of maggma, but related to the problem that a "jointstore" would solve, I have been looking into [GraphQL](https://graphql.org/) (in particular, the [Apollo](https://www.apollographql.com/) implementation) with great interest for providing a unified, single-endpoint API as a layer above a multitude of Stores. I think a basic jointstore could also be useful, though, providing a familiar Mongo/Store API.

montoyjh:
Okay there's one outstanding codacy issue that I think is okay, but this should be ready to go.  I suspect there might still be some bugs, but all of the functionality is tested.

PR #25, opened 2018-05-15
Add read_only flag to Stores, set to read_only by default for source Stores

mkhorton:
@dwinston mentioned this previously, thought it might be useful to enforce good builder hygiene


dwinston:
Great, thanks! Please add a test for this, i.e. assertRaises, and I'll merge.

dwinston:
Sorry, I wasn't clear about the test. I'd like a test that runs a simple builder that tries to write to a source store.

mkhorton:
Np, but I'm not familiar enough with MagicMock to write a builder test, will have to take a look at it another day

shyamd:
I'm closing this for now since this can be achieved with appropriate credentials which are a safer way to ensure ReadOnly and ReadWrite.

PR #30, opened 2018-07-21
[WIP] Improve/refactor examples and move inside maggma namespace

dwinston:
This is WIP, needs tests, etc.

* Runner `run()`s without argument so that serialization of runner is complete, i.e. user knows whether it's an MPI runner or not. This also allows one to inspect a runner's processor before the run, e.g. to customize processor log handling. Also, fixes variable-not-existing error in running with MPI.
* Fix `runner_sample`. Give usage info.
* Make a general `MapBuilder` that takes a montyish-serialized unary function as `process_item` to map source documents to target documents by key. Also, improve the robustness of incremental building via a compound sort on last_updated and store key. `CopyBuilder` is now just a `MapBuilder` with `pydash.identity` as `process_item`.
* Various little fixes/refactors.


dwinston:
Example subclassing of `MapBuilder`:
```python
from maggma.examples.builders import MapBuilder
from pymatgen import Structure
from pymatgen.analysis.xas.spectrum import XANES
from pydash import py_

class XASBuilder(MapBuilder):
    def __init__(self, calcs, xas, **kwargs):
        """MSONable site-specific spectra from calculations.

        Args:
            calcs (Store): XAS calculations with raw output
            xas (Store): output serialized pymatgen XANES objects
        """
        self.calcs = calcs
        self.xas = xas
        self.ufn = msonify_xas
        super().__init__(source=calcs, target=xas, **kwargs)


def msonify_xas(item):
    energy = py_.pluck(item['spectrum'], 0) # (eV)
    intensity = py_.pluck(item['spectrum'], 3) # (mu)
    structure = Structure.from_dict(item['structure'])
    absorption_specie = structure[item['absorbing_atom']].species_string
    edge = "K"
    structure.add_site_property(
        'absorbing_atom', [
            i == item['absorbing_atom']
            for i, _ in enumerate(structure.sites)
        ])
    return {"spectrum": XANES(
        x=energy, y=intensity, structure=structure,
        absorption_specie=absorption_specie, edge=edge,
    ).as_dict()}
```

maggma/examples/builders.py
```diff
@@ -0,0 +1,274 @@
...
+                    for spec in (index['key'] for index in info)))
+    if not all(index_checks):
+        index_warning = (
+            "Missing one or more important indices on stores. "
+            "Performance for large stores may be severely degraded. "
+            "Ensure indices on target.key and "
+            "[(store.lu_field, -1), (store.key, 1)] "
+            "for each of source and target."
+        )
+        logger.warning(index_warning)
```
shyamd: Won't this break when logger is left as the default?

maggma/examples/builders.py
```diff
@@ -0,0 +1,274 @@
...
+        """
+        self.source = source
+        self.target = target
+        self.incremental = incremental
+        self.query = query
+        super().__init__(sources=[source], targets=[target], **kwargs)
+        self.kwargs = kwargs.copy()
+        self.kwargs.update(query=query, incremental=incremental)
+
+    @staticmethod
```
shyamd: why static? Can this and the following static functions be made into properties instead?

maggma/examples/builders.py
```diff
@@ -0,0 +1,274 @@
...
+    def get_items(self):
+        criteria = get_criteria(
+            self.source, self.target, query=self.query,
+            incremental=self.incremental, logger=self.logger)
+        if all(isinstance(entry, str) for entry in self.grouping_properties()):
+            properties = {entry: 1 for entry in self.grouping_properties()}
+            if "_id" not in properties:
+                properties.update({"_id": 0})
+        else:
+            properties = {entry: include for entry, include in self.projection}
```
shyamd: Where does self.projection come from?

maggma/examples/builders.py
dwinston: break? the WARNING log level is higher than info, so a default log level of INFO will pass this. Is the default level higher, e.g. ERROR? I thought INFO was the default.

maggma/examples/builders.py
dwinston: Having it as a method gives a nice error message for free when someone tries to instantiate a subclass without implementing `ufn`. Otherwise, we'd need a `hasattr` check in `__init__` to raise an exception.

maggma/examples/builders.py
shyamd: Right, but the default value for logger is None, and I don't see it being checked or assigned anywhere.

maggma/examples/builders.py
dwinston: oops, a holdover from when I had the class attribute `projection` versus the method `grouping_properties`. Will fix.

dwinston: Done with 89f7c1f

maggma/examples/builders.py
dwinston: From [maggma.builder.Builder](https://github.com/materialsproject/maggma/blob/c4a4ae5db28d6f3eb9dcb0d745873144fa1bf7c0/maggma/builder.py#L35):
```
        self.logger = logging.getLogger(type(self).__name__)
        self.logger.addHandler(logging.NullHandler())
```
The logger is defined and won't "break", but there might be no handler defined, in which case there might not be any output? But calls to logger methods are fine.

PR #32, opened 2018-08-21
Change update_targets log level from debug to exception

mkhorton:
(imo) maggma uses logger.debug slightly too liberally, many times I've had exceptions caught silently, leaving builders in a state where they hang indefinitely

### DIFF
--- a/maggma/runner.py
+++ b/maggma/runner.py
@@ -258,7 +258,7 @@ class MPIProcessor(BaseProcessor):
                     self.update_pbar.update(len(self.data))
                     self.data.clear()
                 except Exception as e:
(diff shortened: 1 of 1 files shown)


mkhorton:
Perhaps uses bare exceptions slightly too liberally too -- my stance on this is that if a builder might fail, that should be handled explicitly by the author inside the builder itself (ie in emmet)?

dwinston:
Unlike with a `SerialProcessor`, which does not catch any potential exception from `builder.update_targets`, letting the exception kill the process will only kill one out of many worker processes. I think the idea to completely halt the builder is fine: in that case, we *do* want to catch the exception in the runner/processor and then do what is needed to stop all workers.

PR #44, opened 2018-10-17
Updates to JointStore and MapBuilder

shyamd:
- Updated JointStore to be able to merge across the lookup to the root level. 
This makes the aggregation look seamless to the person using the JointStore. 
- Modified MapBuilder to get_keys and then chunk those when requesting data. 
This allows the MapBuilder to run in incremental mode by default. 

Todo: Still thinking about how the index functions should be incorporated into MongoStore such that MapBuilder is not calling pymongo functions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,14 @@ language: python
 cache: pip
 services: mongodb
 sudo: required
+addons:
+  apt:
+    sources:
+      - mongodb-upstart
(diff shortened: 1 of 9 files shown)


shyamd:
Ok, This finally works. I think the test is a little mangled, but it was the only way I could get MongoDB 3.6 on the old trusty image that TravisCI seems to use.

PR #54, opened 2019-02-27
Add timeout to MapBuilder, store process time

mkhorton:
So we can modify the BVBuilder to set a max time.

Disabled by default.

### DIFF
--- a/maggma/builders.py
+++ b/maggma/builders.py
@@ -10,6 +10,7 @@ from monty.json import MSONable, MontyDecoder
 from maggma.utils import source_keys_updated, grouper, Timeout
 from time import time
 
+
 class Builder(MSONable, metaclass=ABCMeta):
     """
     Base Builder class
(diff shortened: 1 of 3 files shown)


dwinston:
Could we have `Timeout.seconds` default to `None` instead of `16 * 24 * 60 * 60 / 1e5 == 13.824` seconds? So that a user has to enter the value? We can include "assuming a maximum running time of 1 day for 100,000 items parallelized across 16 cores, you would set this to `16 * 24 * 60 * 60 / 1e5`" as an aid to a user.

mkhorton:
It does default to None in the MapBuilder. If you’re importing Timeout manually you presumably want some kind of non-zero timeout.

dwinston:
Okay, then can the number `13.824` be put somewhere in the docstring to spare any manual importer from having to evaluate that expression?

mkhorton:
Done

PR #56, opened 2019-04-30
Update builder to work with new monty

mkhorton:


dwinston:
Please avoid `KeyError` if "@version" key is absent.

dwinston:
Thanks!

PR #63, opened 2019-11-17
endpoint cluster initial implementation

wuxiaohua1011:
Endpoint cluster initial implementation. Will be dependent on Maggma for data retrieval.

Need to ask Shyam his thoughts on organizing these files.


wuxiaohua1011:
per today's meeting, updated testing to implement a MaterialEndpoint that inherits from EndpointCluster

PR #110, opened 2020-03-04
Amazon S3 store update

munrojm:
AWS profile option added. Also, data will be pulled from S3 index store on query before S3 bucket if possible.


shyamd:
I pushed some clean up commits. Can you add a test for the new behavior?

munrojm:
> I pushed some clean up commits. Can you add a test for the new behavior?

Thanks! Sure, no problem.

PR #111, opened 2020-03-16
MapBuilder retry_failed bug

acrutt:
Bug fix so that retry_failed=True argument of MapBuilder does result in attempting to process failed docs in the target store.


shyamd:
Can you add a test for this so we don't miss it in the future?

acrutt:
I can work on a test but will need more instructions/guidance of what needs to be done since I haven't made one before.

shyamd:
You might actually write tests for your `ProjectionBuilder`. You can model it off the copybuilder or group builder tests. The idea is to make a mock data set, use memory stores to run the builder and make sure all the different functionality works. Feel free to write up something and push the changes here. I can comment on what to do after that. 

https://github.com/materialsproject/maggma/blob/master/tests/builders/test_group_builder.py

PR #114, opened 2020-03-30
Store updates

jmmshn:
Updated various stores for small creature comforts.
- `MongograntStore` now uses the constructor of the parent
- Bumped mongogrant dependency to github version to make this work, will change to pypi version when that becomes availible.
- `S3store` now accepts an optional subdirectory parameter to read and write
- If a sub_dir is defined write it to the metadata.
- Added tests for new `S3Store` features


shyamd:
Why does `mongolike` have such a large drop in testing coverage?

shyamd:
Looks good. Thanks!

PR #115, opened 2020-03-31
small fixes

jmmshn:
enforced s3 key to be a string.
allows sshtunnel to be passed to `MongoStore.connect`


shyamd:
Looks like there is a rogue print statement.

Can you add tests for the `sshtunnel` functionality?

jmmshn:
Ok test added.

PR #128, opened 2020-04-03
Make GridFSStore query check files store first.

munrojm:
This PR makes the query method in the GridFSStore first try and pull data from its mongo `files_store` using the list of `properties` input by the user. 

Also, proper datetime object accommodation is enabled in the S3 index store.

### DIFF
--- a/src/maggma/stores/aws.py
+++ b/src/maggma/stores/aws.py
@@ -268,9 +268,10 @@ class S3Store(Store):
 
             search_docs.append(search_doc.copy())
 
-            search_doc[self.last_updated_field] = str(
-                to_isoformat_ceil_ms(search_doc[self.last_updated_field])
-            )
+            if self.last_updated_field in search_doc:
+                search_doc[self.last_updated_field] = str(
+                    to_isoformat_ceil_ms(search_doc[self.last_updated_field])
+                )
 
             self.s3_bucket.put_object(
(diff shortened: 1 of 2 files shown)


src/maggma/stores/aws.py
```diff
@@ -266,10 +266,15 @@ def update(self, docs: Union[List[Dict], Dict], key: Union[List, str, None] = No
                 search_doc["compression"] = "zlib"
                 data = zlib.compress(data)
 
+            search_docs.append(search_doc.copy())
+
+            search_doc[self.last_updated_field] = str(
+                search_doc[self.last_updated_field]
```
shyamd, line 272: use `maggma.utils.to_isoformat_ceil_ms` to convert the `datetime` object to a `str` so that the format is controlled by maggma.

src/maggma/stores/gridfs.py
```diff
@@ -170,20 +169,38 @@ def query(
...
 
-            try:
-                data = json.loads(data)
-            except Exception:
-                pass
-            yield data
+        for doc in self._files_store.query(
+            criteria=criteria, sort=sort, limit=limit, skip=skip
+        ):
+            if properties is not None and prop_keys.issubset(set(doc.keys())):
```
shyamd, line 181: Do we want this to be on a doc by doc basis or have a store-level property that defines what goes into the metadata?

PR #182, opened 2020-06-02
Reconstruct metadata from index in S3 Store

jmmshn:
Added function to parse index store and populate the metadata of the s3 store.
Converts all the keys to lower case.

Will look into testing with boto3


jmmshn:
@shyamd OK I added the tests.

PR #205, opened 2020-06-22
[WIP] Proper multithreading and msgpack fix

jmmshn:
Changes
- added `msgpack-python` dependency to make unpacking work properly.  Without the dependency `msgpack.upackb` results in byte string for all of the keys and data like`{b'key': b'val'}`
- a test case to reconstruct an MSONable was added to test this capability
- added better handling of threads in aws store: if the updates are only on the Main tread use the aws `resource` and `s3bucket` in the main object, if there are multiple threads each one should have it's own copy of `resource` and `s3bucket` to prevent locking.
- Neet to figure out a test for multithreading write


shyamd:
For the multi-threaded tests, you could ensure that the N>1 is faster than N=1?

jmmshn:
Yeah, that sounds good.

mkhorton:
Have you benchmarked msgpack? I know I was suggesting it, but tested recently and compression wasn't as good as hoped.

jmmshn:
Before zlib compression, our old way of MSONable->json.dumps results in a string that is about 2x bigger than msgpack result.
After zlib they actually result in the file that is around the same size. 
To me, this jsut means that all the compression algorithms are really well optimized.

shyamd:
We did a test before. It might be in the issue thread, if you can find @jmmshn and put a link to it here, that would be good. Basically, I remember it not providing many benefits after compression, but it helped a lot with CPU speed.

jmmshn:
@shyamd I think mocks3 only has one instance so we can't really test multithread.  But the test I have there now should be good enough to check that our code is working and intended

jmmshn:
@mkhorton should we add the `msgpack-python` dependency to Monty? since it really helps with unpacking the data.

src/maggma/stores/aws.py
```diff
@@ -12,9 +13,12 @@
 from maggma.core import Sort, Store
 from maggma.utils import grouper, to_isoformat_ceil_ms
 from monty.dev import deprecated
+from monty.json import MontyDecoder
 from monty.msgpack import default as monty_default
 from monty.msgpack import object_hook as monty_object_hook
 
+thread_local = threading.local()
```
shyamd, line 20: What's the purpose of this module-level variable?

Also, will this cause a problem if we have more than one `S3Store` ?

src/maggma/stores/aws.py
```diff
@@ -174,6 +180,8 @@ def query(
 
                 if doc.get("compression", "") == "zlib":
                     data = zlib.decompress(data)
+                # requires msgpack-python to be installed to fix string encoding problem
+                # https://github.com/msgpack/msgpack/issues/121
```
shyamd, line 182: Don't need this comment if it's made a dependency
```suggestion
```

requirements.txt
```diff
@@ -14,3 +14,4 @@ dnspython==1.16.0
 uvicorn==0.11.5
 sshtunnel==0.1.5
 msgpack==1.0.0
```
shyamd, line 16: Remove this if we're going to have msgpack-python instead
```suggestion
```

src/maggma/stores/aws.py
```diff
@@ -64,6 +68,7 @@ def __init__(
         self.s3 = None  # type: Any
         self.s3_bucket = None  # type: Any
         self.s3_workers = s3_workers
+        self._resources = None
```
shyamd: What is this extra variable here for?

jmmshn: This should be removed

src/maggma/stores/aws.py
jmmshn: Oh yeah that should not be module level

shyamd:
I'm still confused about why you're storing a thread-local resource object. Store's use the MSONable pattern to pickle and unpickle, so each thread should have a new `S3Store`. You should just need to establish a new resource in each `write_doc_to_s3` and that should take care of it no?

jmmshn:
If we have it in `write_doc_to_s3` then the bucket has to be initialized each at each call.
The current implementation only called the constructor once per thread.

PR #241, opened 2020-08-05
accept int as sort keys instead of Sort() in .query() and .groupby()

rkingsbury:
Accept str or int (e.g. "Ascending" or 1) as sort keys instead of requiring `Sort(1)` in `Store.query()`, per previous discussion on Slack.


shyamd:
My only comment is what I wouldn't get rid of the `Sort` enum which is designed to provide some disambiguation. Rather if the sort is provided as an `int` then it should be type-checked against sort to make sure it's a valid number and then used. This could be just converted it to a `Sort` enum. 

The type hint in this situation would change from `Sort` to `Union[Sort,int]`

rkingsbury:
@shyamd the docs test is failing due to a version conflict with `pymdown-extensions`. Is that something I can address?

shyamd:
if you merge in the main repo master it should work now.

Issue #277, opened 2020-09-10
FEATURE: Python file runner


shyamd:
Often we want to parallelize non-module Builders. It would be nice to enable this for at minimum serial and multiprocessing.

shyamd:
Implemented in #278

PR #278, opened 2020-09-11
Add custom source loading

shyamd:
This PR enables adding a custom source file such as `.py` or `.ipynb` file. The only requirements are as follows:
1. The source file has to be in a subfolder from where `mrun` is called. This is a limitation of the pathfinding that has to convert  a module import command into a path
2. The custom python file has to include either a `__builder__` or `__builders__` defined somewhere in the file. This is what gets loaded by `mrun` to run. 

Addresses #277 

Examples:
`mrun -n 2 -v builder_test.py`
`mrun -n 2 -v builder_test.ipynb`
`mrun -n 2 -v main_builders.json builder_test.py`


mkhorton:
This is going to be so useful :)

PR #303, opened 2020-10-08
Small fix to make sure searchable_fields are updated

jmmshn:
# Various fixes
- `newer_in` function in `S3Store` should only look in the the index store of the target
- `searchable_fields` must be included in the `search_keys` in the multiprocessing verion of the update function
- typos in docs


jmmshn:
@munrojm this should fix the speed issues and make the CopyBuilder usable.

src/maggma/stores/aws.py
```diff
@@ -41,7 +41,7 @@ def __init__(
         sub_dir: str = None,
         s3_workers: int = 1,
         key: str = "task_id",
-        searchable_fields: List[str] = [],
+        searchable_fields: List[str] = None,
```
shyamd, line 44:
```suggestion
        searchable_fields: Optional[List[str]] = None,
```

shyamd (commented):
One suggested change.

PR #326, opened 2020-11-12
Direct passing of AWS login to S3Store

jmmshn:
Now allows AWS information to be giving as the s3_profile (should consider renaming the parameter to s3_login).


shyamd:
Looks good!

PR #344, opened 2020-11-26
Setup: msgpack-python was renamed to msgpack

jan-janssen:
https://github.com/msgpack/msgpack-python/issues/280


shyamd:
Thanks for noticing this!

PR #350, opened 2020-12-02
Automatically parse the dbname from the URI

jmmshn:
* Read the database name from the URI, still allow the user-set value to supercede this.


jmmshn:
@shyamd I need to access secrets to add a test for this.

shyamd:
You should be able to use a dummy URI to test if the parsing is working.

jmmshn:
OK awesome!  I thought MongoClient needed to actually connect or something. Guess not

src/maggma/stores/mongolike.py
```diff
@@ -366,15 +366,28 @@ class MongoURIStore(MongoStore):
...
+            with MongoClient(self.uri) as conn:
+                try:
+                    db = conn.get_default_database()
+                except ConfigurationError:
+                    raise ConfigurationError(
+                        "If database name is not supplied, a database must be set in the uri"
+                    )
+            self.database = db.name
+        else:
+            self.database = database
```
shyamd, line 389: does `self.database` even need to be stored if it's not provided; IE, does anything break if `database = None`? Also, this should probably be in `connect` ? Connection on init can be bad.

jmmshn: Ok using the Mongoclient there was a little hacky.
I have changed it.
Technically the code will not break if `database == None` but it will break if `database == None` and no value is found in the URI.
I do think the `database` should be kept since every Mongo-like store must have a database anyways.

PR #381, opened 2021-01-15
update gridfs store

gpetretto:
Currently the GridFSStore does not allow to access data from gridfs collections that were not generated with maggma, since they will not contain the `metadata` value. This PR aims to allow the retrieval of data from a generic gridfs collection.

In its current form however the method returns the bytes extracted from the gridfs and not a dict, as required by the API. In general this could happen if this json.loads fails:
https://github.com/materialsproject/maggma/blob/master/src/maggma/stores/gridfs.py#L209
```
                try:
                    data = json.loads(data)
                except Exception:
                    pass
```

Would you agree to also replace this with:
```
                try:
                    data = json.loads(data)
                except Exception:
                    if not isinstance(data, dict):
                        data = dict(data=data)
```
?


shyamd:
You might also include the key field and the last_updated data from the index. Please update the docstrings to make it clear that is what gets returned.

gpetretto:
@shyamd I hope I interpreted your suggestion correctly.   
I also added a test for this specific use case.

PR #427, opened 2021-04-12
Add Object Hash to S3Store

jmmshn:
hash the data after compression, before insertion.
Then store the hash

added a basic index store reconstruction algorithm.


jmmshn:
@shyamd I think this is ready to merge

PR #435, opened 2021-05-08
Fix aws module import

utf:
Currently, maggma cannot be imported if the botocore library is not installed. botocore is not a core dependency of maggma so this PR fixes the imports.


utf:
The failing test is unrelated to this PR.

PR #464, opened 2021-07-09
Switch from monty to orjson for serialization

munrojm:
This PR switches to using orjson for custom serialization in ReadOnlyResource.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,3 +16,4 @@ uvicorn==0.13.4
 sshtunnel==0.4.0
 msgpack==1.0.2
 msgpack-python==0.5.6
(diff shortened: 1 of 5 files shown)


shyamd:
Why `orjson`? 

It seems silly, but can you add a test for `custom_serialization`. Maybe rename in `object_id_serilaization_helper` to make it clearer what it is doing?

Also, should the flag in `ReadOnlyResource` be called something different like `disable_validation`?

munrojm:
`orjson` is extremely fast and can handle `datetime` object natively while allowing for extensions for things like bson types.

I have renamed things as you have suggested and added the test.

PR #488, opened 2021-09-17
Refactor abstract Drone; add DirectoryDrone

rkingsbury:
The purpose of this PR is to start building out the `Drone` class so that it can be easily used to process data on disk. It has two parts. I've developed this in tandem with a concrete implementation (not part of this PR) that I'm using to process data in my work, and it seems to be working pretty well. Very open to feedback though; I'm very excited to make `Drone` more practical to use, especially for processing experimental data!

## 1. Refactor abstract `Drone`

I studied the example implementation in `SimpleBibDrone` for a long time and felt that it (and the base `Drone` class) could be simplified. I may have missed some of the nuances that motivated the way it was written, but my changes were driven by trying to make `Drone` as consistent with `Builder` as possible. Specifically:

- Eliminated `assimilate()` because it did literally the same thing as `.read()`. If we need `.assimilate()` for compatibility reasons, I can add a deprecation warning and point it to `.read()`
- Eliminated `compute_data` because this is basically equivalent to what `process_item` needs to do. In `Builder`, the approach is to define `process_item` in a very general way (but not make it an `@abstractmethod`) and then allow users to overload it with specific functionality. I copied that approach here.
- Eliminated `compute_record_identifier_key` as a separate required method. There is nothing wrong with having a separate method for this but it doesn't seem to me that it should be enforced in the base class. Ultimately `read()` is the method that needs to generate `RecordIdentifier`, so I think it should be up to each implementation whether the logic of how to generate the corresponding `record_key` should be in `read()` or in a separate method.
- Rename the `store` attribute to `target` for consistency with `Builder`
- Guarantee that the `path` kwarg is converted into a `Path` object even if given as a `str`

## 2. Add `DirectoryDrone` base class

`DirectoryDrone` is intended to work similarly to `MapBuilder`, i.e., to provide a class that allows a user to make a working `Drone` by defining just one `unary_function`. It is intended to be run over a set of directories, where the name of each directory becomes the `RecordIdentifier` key. `unary_function` specifies what to do in each directory.

I could imagine creating an equivalent `FileDrone` that does the same thing but is set up to operate in a single directory full of files, where the filename becomes the record_key.

I need to add tests for `DirectoryDrone` but wanted to get feedback from @shyamd first to make sure I'm heading in the right direction with this.

## Contributor Checklist

   -  [x] refactor `Drone`
   -  [x] add `DirectoryDrone`
- [ ] TODO - I have added tests, or extended existing tests, to cover any new features or bugs fixed in this PR
- [ ] TODO - Add documentation for `DirectoryDrone`


shyamd:
So I've actually had some thoughts about this due to experience with `emmet`

1.) Should we define a base `TaskDocument` here instead and give that the flexible functionality. IE something like a `from_dir` function that will parse the directory into `TaskDocument` subclass.
2.) The  point of `computeRecordIdentifier` was trying to solve was to tell the Drone when it had something it could parse. The idea was that a Drone might be used on a directory that could have multiple sub-documents in it. How does the subclass tell w/e is doing the automation to that it can run on that directory? This is mostly relevant for experimental data where you might have multiple diff documents in the same folder.
3.) Another issue we didn't really solve is the `key`. Maybe this is something we should leave to the `Store` to handle on updates?

rkingsbury:
Thanks @shyamd ! Here are my thoughts on your points

> 1.) Should we define a base `TaskDocument` here instead and give that the flexible functionality. IE something like a `from_dir` function that will parse the directory into `TaskDocument` subclass.

I'm not sure I fully understand how this would work. I guess you're saying that instead of using a `Builder`-like paradigm for `Drone` (with `process_item`, etc.) we'd rely on document models instead? As long as that would still give us all the convenience of `Builder` like multiprocessing, checking when files have changed and need updating, etc. I guess I don't have any objections; I'm just less familiar with this paradigm.

> 2.) The point of `computeRecordIdentifier` was trying to solve was to tell the Drone when it had something it could parse. The idea was that a Drone might be used on a directory that could have multiple sub-documents in it. How does the subclass tell w/e is doing the automation to that it can run on that directory? This is mostly relevant for experimental data where you might have multiple diff documents in the same folder.

In other words, you wanted to be able to include or exclude certain files from processing? I agree this is important; though I think it can be handled by the body of `unary_function`. The approach I took was that the `RecordIdentifier` just gives you a list of all the documents in a directory. It's up to the implementation to pick out the relevant ones and do something with them. I suppose the drawback to this approach is that irrelevant files contribute to the `state_hash` and therefore could trigger unnecessary updates.

What if we just add `include` and `exclude` kwargs to either `read()` or the `__init__` that accept glob patterns or something?

> 3.) Another issue we didn't really solve is the `key`. Maybe this is something we should leave to the `Store` to handle on updates?

The approach I took here was to make the name of each directory the `record_key`. I figured that within a given `parent_dir`, those are required to be unique by the file system. It's not a perfect solution if you're aggregating data from multiple parent directories, but that's not a use case that really had in mind for this class. The workflow I see people (at least, myself) adopting for experimental data is

1. Go to my parent directory full of raw data
2. Process each directory, which may include saving plots or data files in each directory
3. Save results (task documents) to a JSON file in the parent directory (I'm also tinkering with `JSONStore` to make this easier).

So the `record_key` really only needs to work within the context of the one parent directory.  I know this isn't a complete solution for more advanced cases where you want to aggregate data from many directories, but I'd suggest that maybe that's better left for another class. I want to keep this one as simple and straightforward to use as possible.

shyamd:
Yes, I'm saying rather than specialized Drones, we specialize the thing they generate which are `TaskDocument`.

The hierarchy would look like this:
1.) There would still be a `Drone` class in `magma` that is fully implemented already and setup to do all the parallelism necessary: that is for single processing, multi-core processing, and distributed process.

2.) Rather we move the logic to a `AbstractTaskDocument` that is a pedantic data model that then gets specialized. This makes so that we're explicitly defining the specialized thing rather than implicitly describing it via a specialized `Drone` that make some sort of dict object.

As for the computeRecordIdentifier. I realized I completely mixed up the real issue I was trying to solve, so forget what I said above. Instead, imagine a folder with a sample specification JSON, a bunch of XRD measurements, some electrochemical measurements, and maybe a DLS measurement. There are two ways to handle the experimental parsing:
1.) Implement one `ExperimentalTaskDocument` that can handle all of this as a hyper specialized entity.

2.) Implement an `XRDDocument`, `EChemDocument`, and `DLSDocument` that each handle parsing their respective data and know about the sample spec JSON structure to parse that in. The 2nd is sort of important as there are LOTS of different experiments, lots of different manufacturers with different file models and data available. One group alone can have 3 different brands of diffractometers for instance. 

This 2nd option requires some way of looking at that folder, or file, and saying, "I can parse this into a document", which is what `compute_record_identifier_key` was really supposed to be. It clearly got out of hand as I was still mental modeling all of this. 

If we sum this up there would be something like this:

``` python

class AbstractTaskDocument(ABC, MSONable, BaseModel):
    """
    Abstract definition of Task Document
    """

    record_hash: str = Field(description=" A unique ID of the raw data used to make this document") # This was supposed to be the compute record identifier thing, which was a way of checking the DB to see if we hadn't already parsed this data.
    last_updated: datetime = Field(
        default_factory=datetime.utcnow,
        description="Timestamp for this task document was last updateed",
    )

    tags: List[str] = Field([], description="Metadata tags for this task document")
```

Now the real question is what is abstract or implemented by each situation in this model?
1.) One is the function that does the parsing which is your `read` function. 
2.) Is some function that we can use before we run `read` to determine if that even makes sense.
3.) Someway of setting parameters for the read at the `Drone` level .... which I'm at a loss for. Maybe this wasn't a great idea?

This will eventually enable us to do things like run a `Drone` that is parsing VASP, LAMMPS, QChem, CP2K, etc . all at once, or separately if we really want to.

shyamd:
Also, I've been thinking about formalizing the distinction between a `Store` for data and `Store` for documents in the Drone architecture. That is a `Drone` has two targets: one place for `TaskDocuments` and one place for `DataBlobs` or something. This could be CHGCARs or w/e, but the idea is that its binary data that is not intended to be searched, indexed etc. Again, this idea tends to lean to the `AbstractDrone` hierarchy, but we could also make `DataBlob` a type and put that in a specialized key in `AbstractTaskDocument`. More ramblings.

shyamd:
Just a a thought for the issue of `settings` for the specialized `AbstractTaskDocument`. We could use the universal settings file archetype in `emmet`. Basically there is a base `Settings` class that any `TaskDocument` can implement its own customized version. The import of this happens in the base, and the specialized versions just ignore things that are not for them. This could cause some clashing, but it is an option. MoreRamblingsRUs.

rkingsbury:
> 1.) There would still be a `Drone` class in `magma` that is fully implemented already and setup to do all the parallelism necessary: that is for single processing, multi-core processing, and distributed process.

I see. Are there any technical reasons why it would be easier to put parallelizable work into `process_items` as opposed to a `TaskDocument.from_directory()`?

The other reason I hesitate about this approach is that it creates another paradigm for users to learn. One of the reasons I find `maggma` appealing to use is that it has relatively few high-level interfaces to learn (`Store` and `Builder` and hopefully `Drone` which is similar). Taking the `TaskDocument` approach means that there's a new object in the mix that is conceptually very different.

> 2.) Rather we move the logic to a `AbstractTaskDocument` that is a pedantic data model that then gets specialized. This makes so that we're explicitly defining the specialized thing rather than implicitly describing it via a specialized `Drone` that make some sort of dict object.

I see the value in this.  I do want to point out that generating dicts or taskdocs is only half of the reason that I want to use `Drone` though. There are also many cases where I just have a lot of in-place data processing to do (to take the MD example, maybe I'm computing RDFs, analyzing solvation or transport properties, saving plots and data files in the local directory, etc.). For this kind of compute-heavy analysis I want the `Drone` / `Builder` infrastructure to help me multiprocess, pause and resume, only update when needed, etc. This is valuable even if you don't generate a `TaskDocument` at all.

I recognize this might not be the paradigm `maggma` is built around - we are used to putting _everything_ into databases, and I do like that approach. But I think there's an opportunity to make the `Builder` infrastructure really useful to a lot of users (i.e. experimentalists) who aren't likely to make the jump into a full database approach to managing data.

None of this precludes putting the specialization in `AbstractTaskDocument`, but maybe it's less natural to write in-place data processing steps in a task document object if you aren't actually interested in that object?

> 2.) Implement an `XRDDocument`, `EChemDocument`, and `DLSDocument` that each handle parsing their respective data and know about the sample spec JSON structure to parse that in. The 2nd is sort of important as there are LOTS of different experiments, lots of different manufacturers with different file models and data available. One group alone can have 3 different brands of diffractometers for instance.

Agreed. I think you'd probably want to have specific `Document` classes. Each one reads the directory looking for specific file names or extension and only processes those. I don't think the average experimenter has the discipline to keep data from a single analysis in a single directory all the time. That said, I still think using the directory name as the `RecordIdentifier` key could make sense and would be intuitive.

> 
> Now the real question is what is abstract or implemented by each situation in this model?
> 1.) One is the function that does the parsing which is your `read` function.
> 2.) Is some function that we can use before we run `read` to determine if that even makes sense.
> 3.) Someway of setting parameters for the read at the `Drone` level .... which I'm at a loss for. Maybe this wasn't a great idea?

To me it seems natural to combine 1) and 2), but I think I'm understanding why you had a separate `computeRecordIdentifierKey` method now. For example, if I have 20 files in a directory, and I want an `ExcelDocument` to parse the one called 'results.xlsx' and nothing else, then my `RecordIdentifier` really only needs to include that one document and no others.

Even so, I feel like this type of logic could just be integrated together with the `read()` logic under `TaskDocument.from_directory()`.

As for 3), I'll paste an example in the next post that I've been using these last few weeks. This follows the pattern of a concrete implementation of `MapBuilder`.

> 
> This will eventually enable us to do things like run a `Drone` that is parsing VASP, LAMMPS, QChem, CP2K, etc . all at once, or separately if we really want to.

Sounds great, but is this really necessary? I worry that it will lead to us over engineering this and making it needlessly complex for simpler, more common use cases.

rkingsbury:
Example concrete implementation of `DirectoryDrone` that passes kwargs to `unary_function`

```
class RDFDrone(DirectoryDrone):
    """
    Analyze radiual distribution functions (RDFs) from a directory of LAMMPS trajectories.
    Each trajectory is assumed to reside in a separate subdirectory inside the parent
    directory.
    """
    
    def __init__(self, path,
                 target,
                 select_dict: Dict,
                 rdf_kwargs: Dict = {},
                 rdf_run_kwargs: Dict = {},
                 split: int = 1,
                 write_plot: bool = True,
                 write_data: bool = True, **kwargs):
        """
        Args:
            path,
            target,
            select_dict: dictionary of species label: MDAnalysis selection string
                that specifies the distinct species in the trajectory. RDFs will be
                generated for all pairwise combinations of species in select_dict
                e.g. {"cation": "type 3",
                      "anion": "type 4",
                      "water": "type 2"}
            split: Number of time periods over which to compute RDFs. The trajectory frames
                will be divided into this many sub-trajectories and a separate RDF will be
                computed for each. This can be used as a way of verifying that a trajectory
                has converged with respect to solvation structure. For example, split=2 will
                compute an RDF for the first half and the second half of the trajectory.
                split=1 (Default) computes the RDF for the entire trajectory.
                
                Multiple RDFs will be overlayed onto one plot if write_plot is True.
                Note that if 'start' and 'stop' are passed to rdf_run_kwargs, the split
                will be performed only on the frames between start and stop, not on the entire
                trajectory.
            rdf_kwargs: kwargs passed to MDAnalysis.analysis.rdf.InterRDF on instantiation.
                For example, you can set {"range": (0.1, 20)} to modify the cutoff distances
                used in the calculation.
            rdf_run_kwargs: kwargs passed to MDAnalysis.analysis.rdf.InterRDF.run(), 
                e.g. {"start":2, "stop": 100, "step": 10}. By default every frame
                of the entire trajectory is analyzed. Settting 'step' to a value larger
                than 1 will speed up the analysis by skipping frames
            write_plot: Whether to save a .png of the RDF plot in the calculation
                directory (Default: True)
            write_data: Whether to save the RDF data as a binary .npy array in the
                calculation directory. This can save time if you want to re-load the
                data for subsequent analysis without having to re-run the RDF.
                (Default: True)
        """
        self.select_dict = select_dict
        self.rdf_kwargs = rdf_kwargs
        self.rdf_run_kwargs = rdf_run_kwargs
        self.split=split
        self.write_plot = write_plot
        self.write_data = write_data
        super().__init__(path, target, **kwargs)
        
    
    def unary_function(self, item: RecordIdentifier):
        """
        Compute radial distribution functions
        """
        # construct MDAnalyis trajecory
        data_file = [d for d in item.documents if d.name == 'md.data'][0]
        dcd_file = [d for d in item.documents if '_production.dcd' in d.name][0]
        traj = Universe(str(data_file.path),
                        str(dcd_file.path))
        
        # create atom groups
        atom_groups = {}
        pairwise_combos = [] # list of tuples specifying pairs of species to analyze
        for k,v in self.select_dict.items():
            atom_groups[k] = traj.select_atoms(v)
            # add self pairs to the list of combos to evaluate
            pairwise_combos.append((k,k))
        
        # add all pairwise combinations
        pairwise_combos.extend(list(itertools.combinations(self.select_dict.keys(), 2)))
        
        # dict to collect results
        out = {"rdfs": {}}
      
        ### LOTS OF LOGIC TO COMPUTE RDFs

       self.logger.info(f'Finished computing {pair[0]}-{pair[1]} RDF for {item.record_key}')
        
       return out
```

shyamd:
The point of putting the  work in `process_items` rather than explicitly parallelizing in `read_directory` is that `maggma` can then use w/e parallelism is available, IE single core, multi-core, multi-node. It's not easy to figure out how to do some of these things and `maggma` mostly takes care of that if you follow its rules. This would also eliminate the need to specialize `Drone`. IE, people would just implement `TaskDocument` classes and feed them as a configuration to `Drone`. Again, this is not a fully fleshed out concept, just something I threw out there to see if it made more sense. 
 
I'm mostly onboard with this PR. I still think the `Drone` model will morph more into `Document` model more like in `emmet`. This can happen gradually via a specialized `Drone`. Do we still need the `RecordIdentifier`? It feels so.... clunky, but I can't mentally get rid of it. Also, how is your implementation dealing with different directories for different calcs? Is `unary_function` getting all of the files in all directories? If that's the case you're losing parallelism. `get_items` needs some way of dispatching different directories to different process_items, which is another reason for `compute_record_identifier`. 

I also agree that not everyone wants to use a DB. The `Store` is just a bunch of data in a DB like model, so there's no reason it can't be backed by a  Pandas DataFrame for instance, where a Drone + Builders just run a series of operations. The DataFrame will break down if multi-node processing, which is where a DB is necessary.  It's this kind of nuance that I'm trying to simplify. Normally, this would take a whole separate bit of code, but since the `Store` interface is mutual, you would just switch from a DataFrame `Store` to a DB `Store` of some type. In your `RDFDrone` example, I would split this into two objects: one `LAMMPSDrone` and a `RDFBuilder`, where the `target` of the drone and the `source` of the builder were one consistent `MemoryStore`. I realized that currently that doesn't work because `MemoryStore` always makes a new `mongomock` collection which prevents stringing together builders this way.

rkingsbury:
Closing based on discussion with @munrojm . I'll resurrect this as a type of `FilesStore` with similar functionality in a separate PR.

PR #493, opened 2021-09-29
Change to S3Store serialization behavior in update() and other Mongolike Store changes

sivonxay:
1. Change the behavior of s3/minio stores, so metadata can be queried by its original type. 
2. Added from_launchpad_file for GridFS Store
3. Added kwargs to MongoStore.from_launchpad_file

### Description of 1:

* Store metadata as original type in index collection. For example, when inserting the following document into an s3store
  ```
  doc = {'formula_pretty': 'Si', 'nsites': 10, 'composition': {'Si': 1}}
  s3store.update(doc, key='formula_pretty')
  ```
  will be stored in the index with all keys and values converted to strings
  ```
  doc = {'formula_pretty': 'Si', 'nsites': '10', 'composition': "{'Si': 1}"}
  ```

* This also fixes an issue with having non-string types passed to s3_bucket.put_object() in the search_doc. Don't fully understand the issue, but it's possible one of the items that was later was not of type str.

### DIFF
--- a/src/maggma/stores/aws.py
+++ b/src/maggma/stores/aws.py
@@ -34,18 +34,18 @@ class S3Store(Store):
     """
 
     def __init__(
-        self,
-        index: Store,
-        bucket: str,
-        s3_profile: Union[str, dict] = None,
-        compress: bool = False,
-        endpoint_url: str = None,
-        sub_dir: str = None,
-        s3_workers: int = 1,
-        key: str = "fs_id",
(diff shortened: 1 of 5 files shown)


munrojm:
Thanks for your work on this @sivonxay. Are you looking to have it merged now?

sivonxay:
Not urgent, but if all looks good to you, yes please!

PR #514, opened 2021-12-01
Added on-disk MongoDB compatible MontyStore

utf:
In my opinion, one missing component of the maggma stores is a persistent on disk store that can be accessed simultaneously from multiple python processes. Think an almost fully formed MongoDB instance but hosted locally on the file system. I think this is what #507 is hinting towards.

In this PR I've implemented a store based on the [MontyDB](https://github.com/davidlatwe/montydb) backend. This provides a MongoDB interface to several on disk stores including:

- sqlite: Uses an sqlite database to store documents.
- lightning: Uses Lightning Memory-Mapped Database (LMDB) for storage. This can provide fast read and write times but requires lmdb to be installed (in most cases this can be achieved using ``pip install lmdb``).
- flatfile: Uses a system of flat json files. This is not recommended as multiple simultaneous connections to the store will not work correctly.

The data is stored in a folder on disk controlled by the ``database_path`` option. Multiple databases and collections are stored as different files.

I've added full tests and MontyDB as an optional requirement. @davidwaroquiers you may be interested in this.

### DIFF
--- a/requirements-optional.txt
+++ b/requirements-optional.txt
@@ -5,3 +5,4 @@ IPython==7.16.1;python_version<"3.7"
 IPython==7.25.0;python_version>"3.6"
 nbformat==5.1.3
 regex==2020.11.13
+montydb==2.3.12
--- a/setup.py
+++ b/setup.py
@@ -47,6 +47,7 @@ setup(
     extras_require={
         "vault": ["hvac>=0.9.5"],
         "S3": ["boto3>=1.14.56"],
+        "montydb": ["monty>=2.3.12"],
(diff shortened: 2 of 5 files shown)


shyamd:
Is there any reason to not just replace the monogmock `MemoryStore` based implementations with this `montydb` variant?

utf:
Maybe. The existing memory store is based on mongomock which I think is a more feature complete replica of pymongo. For example, it implements bulk_write whereas MontyDB does not.

munrojm:
@utf I agree with your sentiment, and like this implementation. Are you good for me to merge?

utf:
Yeah all good from me. I've been using this and seems to work well.

munrojm:
Sounds good!

PR #567, opened 2022-02-07
Add authsource option for mongo and gridfs stores

utf:
As the title says.

By default the authsource is the database specified (replicating the existing behaviour).


mkhorton:
Thanks @utf

PR #573, opened 2022-03-03
added mongoclient_kwarg option

jmmshn:
Allows kwargs to be passed to `MongoClient` in `MongoStore`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
     - uses: actions/setup-python@v3
       with:
@@ -36,7 +36,7 @@ jobs:
     - deploy
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
     - uses: actions/setup-python@v3
       with:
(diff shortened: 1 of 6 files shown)


munrojm:
Is there a reason for `kwargs` instead of explicitly defining extra parameters? What was the issue you ran into?

jmmshn:
I have the following in my launchpad file for my atomate1 setup.
```
mongoclient_kwargs:
    ssl: true
    tlsAllowInvalidCertificates: true
```
and I'm just trying to replicate the behavior for atomate2 and I think there are lots of kwargs for MongoClient we are not using.

munrojm:
The `kwargs` that is already there is only used when calling `__init__` of `Store`. That method can only take in four arguments: `key`, ` last_updated_field`, `last_updated_type`, and ` validator`. What about just using the existing `kwargs`, and simply checking for those four arguments and passing them explicitly to `Store.__init__` if they exist? You can then pass the rest to `MongoClient`, and make it clear in the docstring details that users can pass things to `MongoClient`. 

Just a suggestion, but it would make things a bit cleaner.

jmmshn:
Actually, I'm realizing that allowing the code to do this logic is going to cause problems for things downstream like `GridFSStore` and other things that inherit from Store like during construction of JobStores in jobflow, there can be a lot of ambiguities. I think we have to be explicit here and use `mongocleint_kwargs` which makes debugging problems a lot easier.

PR #574, opened 2022-03-04
change cleint -> resource in aws tests

jmmshn:
http://docs.getmoto.org/en/latest/docs/getting_started.html

Similar to a fix to this PR on automate:
https://github.com/hackingmaterials/atomate/pull/749


munrojm:
Does explicitly specifying `us-east-1` here impact anything?

jmmshn:
no but I think moto wants you to do this because sometimes missing regions gives warnings and defaults to us-east

PR #575, opened 2022-03-04
mongoclient_kwargs

jmmshn:
Allows additional kwargs to be passed to `MongoClient` in different stores that contain `MongoClient` inits.

### DIFF
--- a/src/maggma/stores/mongolike.py
+++ b/src/maggma/stores/mongolike.py
@@ -122,6 +122,7 @@ class MongoStore(Store):
         ssh_tunnel: Optional[SSHTunnel] = None,
         safe_update: bool = False,
         auth_source: Optional[str] = None,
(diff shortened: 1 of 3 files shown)


src/maggma/stores/mongolike.py
```diff
@@ -564,7 +567,7 @@ def connect(self, force_reset: bool = False):
         Connect to the source data
         """
         if self._coll is None or force_reset:  # pragma: no cover
-            conn = MongoClient(self.uri)
+            conn = MongoClient(self.uri, self.mongodb_client_kwargs)
```
munrojm, line 570: I think this needs `**self.mongoclient_kwargs`, no?

jmmshn: oops! fixed

PR #595, opened 2022-03-08
added localhost test for MongoURIStore

jmmshn:
The constructor for MongoURIStore bypasses the MongoStore constructor so the `mongoclient_kwarg` has to be explicitly added here.
Added test to catch these next time


jmmshn:
Found this while doing a PR on atomate:
https://github.com/hackingmaterials/atomate/pull/751

PR #619, opened 2022-04-06
FileStore: a Store for files on disk

rkingsbury:
This WIP PR is an attempt to facilitate using `maggma` infrastructure to process files on disk (e.g. experimental data files). It is a follow up to the closed #488. ~~This is in a very early state and intended mainly to facilitate discussion.~~

~~**UPDATED 2022-04-11**~~
~~**UPDATED 2022-04-12**~~
~~**UPDATED 2022-04-13**~~
~~**UPDATED 2022-04-18**~~
**UPDATED 2022-04-25**

Based on discussion with @munrojm , the idea here is to define a `FileStore` that provides a `maggma` style interface a to a directory full of files, making it possible to run `Builder` directly on data files and not just on mongo documents. That way, you can process files without having to write a separate `Drone` class for parsing. Rather, you can point a `Builder` at a `FileStore`, use the `FileStore` to retrieve the relevant files, and then define whatever work needs to happen in `process_items` of a `Builder`.

~~Right now the `FileStore` is initialized with a parent directory, and each _subdirectory_ constitutes one item or record in the store. Each record contains a list of the files in that subdirectory and keeps track of their modification times.~~ There is an internal `MemoryStore` that tracks all the file metadata.

Here is my thought process behind the current implementation:
1. It seems to me that `update` and `remove_docs` should actually create and delete files on disk if we are to adhere strongly to the `Store` paradigm with respect to how documents work (i.e., the files on disk that constitute the `Store` must ALWAYS be in sync with contents of the internal `MemoryStore`. We don't want to delete some records from the `MemoryStore` without changing the corresponding files).
2. If we adhere to Point # 1, then the only way to add additional metadata to a record is to write a new file. So for example, if I added a field to a record, as soon as I call `update` the `FileStore` should simultaneously 1) add the field to the mongo document in the internal `MemoryStore` and 2) create a .json file in the respective directory that includes the metadata. 
3. ~~Point # 2 is the reason that I've built `FileStore` to use _directories_ as the items, rather than individual files. Directories are guaranteed to have unique names and they provide a place to store additional metadata that one might want to add.~~
4. ~~It would also be useful to have a version of this class that can operate on a single directory full of individual files. In this case, the metadata for all records would have to be written to a single .json file.~~
5. ~~Maybe it would make sense to define a base `FileStore` based on point # 4 and an inherited class that is further customized  for data organized into directories.~~
6. I've revised the `FileStore` so that each item represents a single file. Metadata is written automatically to a .json file in root directory of the `FileStore`.

Basic usage and concepts are as follows:

- On `.connect()` `FileStore` iterates through all files in its base directory and create a `FileRecord` object from each. `max_depth` and `file_filters` (`fnmatch` patterns) can be passed as kwargs to restrict which files are indexed.
- On `.connect()`, if the store is not read-only, `FileStore` will create and connect to an internal `JSONStore` that creates or reads a .json file of metadata in the base directory
- ~~`FileRecord` objects record the name, parent directory, size, hash, and last modified time of the file and populate that into the internal `MemoryStore`~~
- The internal `MemoryStore` is populated with dicts containing the name, full path, parent directory, size, hash, and last modified time of the file. This is accomplished via the `.read()` method.
- Files are uniquely identified by a `file_id`, which is computed by `read()` as the hash of the file path _relative to the FileStore base directory_. This relative path is guaranteed to be unique by the file system. By using the relative path instead of the absolute path, we make it possible to move the entire `FileStore` to a new location on disk without changing `file_id`. NOTE: we considered adding the file creation time to the hash that constitutes the `file_id`. However I decided against this because its value may differ across platforms. E.g. the `pathlib.stat().st_ctime` attribute returns different values on Windows vs. Unix/Mac - see [here](https://pynative.com/python-file-creation-modification-datetime/#h-pathlib-module-to-get-the-creation-and-modification-datetime-of-a-file))
- On `.update()`, anything in the document _except_ the keys populated by `.read()` will be written out to the .json file (and read back in next time you connect to the store). Only items with extra keys are written to the JSON (i.e., if you have 10 items in the store but add metadata to just one, only the one item will be written to the JSON). The purpose of this behavior is to prevent any duplication of data. The `file_id` (or whatever `self.key` is) and `path` are retained in the JSON file to make each metadata record manually identifiable
- In the event that there are metadata records in the JSON file that no longer match files on disk, those records will be added to the FileStore and marked with `{"orphan":True}`. This can happen if, for example, you init a `FileStore` and later delete a file, or if you init the store with the default arguments but later restrict the file selection with `max_depth` or `file_filters`. The goal with this behavior is to preserve all metadata the user may have added and prevent data loss.
- There is an `include_orphans` kwarg you can set on init that defines whether or not orphaned metadata records will be returned in queries
- `remove_docs` deletes files assuming the store is not read only. It has an additional guard argument `confirm` which must be set to the non-default value `True` for the method to actually do anything.
- `query` will attempt to read the actual contents of a file up to a certain size limit (which can be adjusted via kwarg). The contents are added under a `contents` key in the returned item
- Here is an example returned document (a file from the `test_files` directory)

```
{'_id': ObjectId('625e581113cef6275a992abe'),
 'name': 'input.in',
 'path': '/home/ryan/miniconda3/envs/md/code/maggma/tests/test_files/file_store_test/calculation1/input.in',
 'parent': 'calculation1',
 'size': 90,
 'file_id': '2d12e9803fa0c6eaffb065c8dc3cf4fe',
 'last_updated': datetime.datetime(2022, 4, 19, 5, 23, 54, 109000),
 'hash': 'd42c9ff24dc2fde99ed831ec767bd3fb',
 'orphan': False,
 'contents': 'This is the file named input.in\nIn directory calculation1\nin the FileStore test directory.'}
```

Eager to hear thoughts!

## Contributor Checklist

   -  [x] Create a `FileStore` class that uses files on disk as the underlying data storage medium
   - [x] implement `remove_docs` method
   - [x] add a convenience method to assign metadata by parsing the file or directory name
   - [ ] Refactor / streamline
   - [ ] Write documentation and update docstrings

## Possible future enhancements (separate PRs)
- [ ] Integrate compression somehow
- [ ] Replace `mongomock` with a faster in-memory alternative

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -85,7 +85,7 @@ jobs:
       run: |
         pip install -e .
         pytest --cov=maggma --cov-report=xml
-    - uses: codecov/codecov-action@v3.0.0
+    - uses: codecov/codecov-action@v3.1.0
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
         file: ./coverage.xml
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,21 @@
 # Changelog
 
+## [v0.46.1](https://github.com/materialsproject/maggma/tree/v0.46.1) (2022-04-21)
+
+[Full Changelog](https://github.com/materialsproject/maggma/compare/v0.46.0...v0.46.1)
+
+**Merged pull requests:**
+
+- Prefix `fields` input for read resource key endpoint [\#636](https://github.com/materialsproject/maggma/pull/636) ([munrojm](https://github.com/munrojm))
+
+## [v0.46.0](https://github.com/materialsproject/maggma/tree/v0.46.0) (2022-04-19)
+
+[Full Changelog](https://github.com/materialsproject/maggma/compare/v0.45.1...v0.46.0)
+
+**Merged pull requests:**
+
+- S3 store and resource additions [\#635](https://github.com/materialsproject/maggma/pull/635) ([munrojm](https://github.com/munrojm))
+
 ## [v0.45.1](https://github.com/materialsproject/maggma/tree/v0.45.1) (2022-04-18)
 
 [Full Changelog](https://github.com/materialsproject/maggma/compare/v0.45.0...v0.45.1)
@@ -50,15 +66,11 @@
 
 ## [v0.44.1](https://github.com/materialsproject/maggma/tree/v0.44.1) (2022-03-08)
 
-[Full Changelog](https://github.com/materialsproject/maggma/compare/v0.42.0...v0.44.1)
+[Full Changelog](https://github.com/materialsproject/maggma/compare/v0.43.0...v0.44.1)
 
 **Merged pull requests:**
 
 - added localhost test for MongoURIStore [\#595](https://github.com/materialsproject/maggma/pull/595) ([jmmshn](https://github.com/jmmshn))
-
-## [v0.42.0](https://github.com/materialsproject/maggma/tree/v0.42.0) (2022-03-07)
-
-[Full Changelog](https://github.com/materialsproject/maggma/compare/v0.43.0...v0.42.0)
 
 ## [v0.43.0](https://github.com/materialsproject/maggma/tree/v0.43.0) (2022-03-07)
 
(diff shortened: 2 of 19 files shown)


rkingsbury:
Hi @munrojm I'm going to take the `FileStore` PR out of WIP status because I think it's basically feature-complete, tested, and working well. I suspect there are some ways to refactor the "guts" of this class to be a little cleaner, reduce redundancy from parent classes, etc. I feel confident now that as long as the tests continue to pass we can rework the internals as desired. If you have any feedback in that regard (or with respect to function and kwarg naming, organization, etc.) please let me know.

Also, I would like to add a thorough documentation page before merge but I thought I'd wait for feedback from you on the interface and implementation before I write that.

Lastly, I will note that as we discussed before, the slowness of `mongomock` is a bit annoying, even when using `FileStore` on a directory of only ~500 small text files. Hopefully we can find a faster alternative someday.

munrojm:
Hi @rkingsbury, that sounds great! Thanks again for all of your hard work on this. I will set aside some time tomorrow to do a detailed line-by-line and add my comments.

PR #621, opened 2022-04-11
MemoryStore: fix groupby ignoring properties

rkingsbury:
In the course of developing `FileStore` (#619 ) I discovered that `MemoryStore.groupby()` was not correctly passing the `properties` argument, making it impossible to project fields in the returned data. This small PR fixes that behavior.

## Contributor Checklist

   -  [x] Fix broken `MemoryStore.groupby`

### DIFF
--- a/requirements-testing.txt
+++ b/requirements-testing.txt
@@ -1,4 +1,4 @@
-pre-commit==2.17.0
+pre-commit==2.18.1
 pytest==7.1.1
 pytest-asyncio==0.18.2
 pytest-cov==3.0.0
--- a/src/maggma/stores/mongolike.py
+++ b/src/maggma/stores/mongolike.py
@@ -638,9 +638,15 @@ class MemoryStore(MongoStore):
             generator returning tuples of (key, list of elemnts)
         """
         keys = keys if isinstance(keys, list) else [keys]
+
+        if properties is None:
+            properties = []
+        if isinstance(properties, dict):
+            properties = list(properties.keys())
+
         data = [
             doc
-            for doc in self.query(properties=keys, criteria=criteria)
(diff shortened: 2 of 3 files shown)


munrojm:
Thanks for catching this @rkingsbury. My vote would be to by default pass `properties` instead of  `keys + properties`. This would allow someone to leave the properties field blank if they wanted the complete document.

rkingsbury:
> Thanks for catching this @rkingsbury. My vote would be to by default pass `properties` instead of `keys + properties`. This would allow someone to leave the properties field blank if they wanted the complete document.

That makes sense, but I think `keys` is required here. If I remove `keys` and just pass `properties=properties`, the associated test fails. Also in `MongoStore.groupby` there is [a line](https://github.com/materialsproject/maggma/blob/0c1ea7facbd97d53dc8df7b22dbbafc00f2eb92f/src/maggma/stores/mongolike.py#L293) that adds `keys` to whatever properties the user passes (that's actually what tipped me off to the cause of this problem). Let me know if you think there's a better solution though.

munrojm:
Ah, okay that is fair. Thanks for clarifying. I'll merge and patch release.

rkingsbury:
Hi @munrojm a follow up to this - I've since discovered that if you do NOT pass `properties`, then you don't get any returned data from `MemoryStore.groupby()`. Is that a bug or intended behavior? Based on the behavior of `query` in which omitting `properties` returns the whole document I thought the same would be true for `groupby`. 

I can work on a fix if this is indeed a bug. Thanks!

munrojm:
Other stores do indeed to return the full data if no properties are passed. I think we should fix it.

PR #624, opened 2022-04-11
JSONStore: write file on init, add descriptive KeyError, add tests

rkingsbury:
This PR makes a few enhancements to `JSONStore`.

## Contributor Checklist

   -  [x] Allow `JSONStore` to create an empty .json file when `file_writeable=True` and the path given does not already exist
   -  [x] Raies a more descriptive error when a KeyError occurs on `.connect()`. This is helpful because when working with `JSONStore`, it is easy to init with a non-default key field and then forget that you did that when you try to reconnect to the store.
   - [x] ~~add a helper class to make it possible to write `datetime` objects to JSON. If there is already a solution to this somewhere in `maggma` please advise and I'll modify accordingly.~~
   - [x] migrate `JSONStore` from `json` to `orjson` 
   - [x] Add tests for the above and for an additional warning about multiple paths that was not previously tested

### DIFF
--- a/src/maggma/stores/mongolike.py
+++ b/src/maggma/stores/mongolike.py
@@ -5,7 +5,6 @@ Stores are a default access pattern to data and provide
 various utilities
 """
 
-import json
 from pathlib import Path
 from datetime import datetime
 import yaml
@@ -14,6 +13,7 @@ from socket import socket
 from typing import Any, Dict, Iterator, List, Optional, Tuple, Union
 
 import mongomock
+import orjson
 from monty.dev import requires
 from monty.io import zopen
 from monty.json import MSONable, jsanitize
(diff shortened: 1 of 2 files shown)


src/maggma/stores/mongolike.py
```diff
@@ -750,7 +766,7 @@ def update_json_file(self):
             data = [d for d in self.query()]
             for d in data:
                 d.pop("_id")
-            json.dump(data, f)
+            json.dump(data, f, default=json_serial)
```
munrojm, line 769: Might make sense to use ORJSON here. It can natively serialize datetime objects and would be faster. It is also already a dependency of maggma.

munrojm:
Added one comment. Looks good to me other than that.

src/maggma/stores/mongolike.py
rkingsbury: Thanks, done! I was not familiar with `orjson` but it looks like a good move.

rkingsbury:
Thanks @munrojm ! What would you think about renaming the `JSONStore`  `file_writeable` kwarg to `read_only`? I feel like that would be easier to understand and consistent with what we're thinking for `FileStore`. Could put some logic in to retain backward compatibility if that's a concern.

rkingsbury:
> think that is a good idea. If we do, I think we should keep backwards compatibility as you say and issue a deprecation warning.

See #625

PR #625, opened 2022-04-12
JSONStore: file_writable -> read_only

rkingsbury:
This PR adds a `read_only` kwarg to `JSONStore` to replace `file_writable`.

## Contributor Checklist

   -  [x] add read_only
   - [x] add backwards compatibility logic to continue supporting file_writable

### DIFF
--- a/src/maggma/stores/mongolike.py
+++ b/src/maggma/stores/mongolike.py
@@ -6,10 +6,10 @@ various utilities
 """
 
 from pathlib import Path
-from datetime import datetime
 import yaml
 from itertools import chain, groupby
 from socket import socket
+import warnings
 from typing import Any, Dict, Iterator, List, Optional, Tuple, Union
 
 import mongomock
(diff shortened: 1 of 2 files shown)


rkingsbury:
> I think we should maybe only keep `read_only` as an explicit argument that is passed. We can then absorb any mention of `file_writable` into `**kwargs`. Do you have an opinion on that?

Just to make sure I understand what you're saying, do you mean, e.g. 

```
class JSONStore(MemoryStore):
    """
    A Store for access to a single or multiple JSON files
    """

    def __init__(
        self,
        paths: Union[str, List[str]],
        read_only: bool,
        file_writable: Optional[bool] = None,
        **kwargs,
    ):
```

in other words, the user always has to specify `read_only` ?

munrojm:
Yeah, exactly.

rkingsbury:
OK, will this cause compatibility problems though? Because if we're trying to support legacy code that says, e.g. `js=JSONStore(<filename>)` that would start to fail since it doesn't specify `read_only`.

munrojm:
Oh, sorry for not being clear. It can be optional, I just wanted to suggest changing how `file_writable` is handled so it isn't explicitly in the function signature anymore.

PR #626, opened 2022-04-18
minor bug fix in remove_docs

jmmshn:
obj might not be a string


munrojm:
Thanks @jmmshn! Would you be able to alter the relevant s3 store test so that the line is covered?

jmmshn:
@munrojm coverage passed

munrojm:
Sounds good, thanks!

PR #669, opened 2022-05-27
Proposal: remove Drone class

rkingsbury:
The recently merged [`FileStore`](#619 ) provides a way to connect local files on disk to `maggma` `Store` and `Builder`, and hence serves the purpose that `Drone` was originally envisioned for without requiring an additional interface. 

Given that, and that there is still an alternate implementation of `Drone` living in pymatgen, I propose that we remove `Drone` from maggma to minimize user confusion. `FileStore` can be the "maggmonic" way of interfacing local files with a data store, and `Drone` can be an alternate way to accomplish that tasks for users that prefer other codes.

The maggma implementation of `Drone` was functional but in a pretty early stage of development, documentation, and maturity anyway, and I feel fairly confident that no one was actually using it for anything (except perhaps for me, in what later became `FileStore` - see #488 ).

This PR removes `Drone` completely. Obviously if there are concerns about compatibility we could use a deprecation warning to make a smoother transition.

Happy to hear comments, or feel free to close if you want to retain `Drone`. I recognize that leaving it in the codebase doesn't really hurt anything, but I'm concerned that it's presence (especially in documentation) will confuse users about the preferred way to get local files into `Store`.


rkingsbury:
Just following up on this one as well. I see a thumbs up from @shyamd so I take that to mean he's on board with getting ride of `Drone`.  I think the question is are we OK stripping out `Drone` immediately (as done here), or do we feel its important to put in some deprecation warnings for a few versions before we do so? 

I would generally always advocate deprecation warnings before a breaking change like this. In this particular instance I feel OK removing `Drone` in one fell swoop only because I _think_ it was never used by anyone. But I'd be happy to modify this such that we keep the class with a `DeprecationWarning` and just remove mentions of it from the docs.

munrojm:
I'm happy to just remove it as well. Like you say, I am not sure anyone is using it for anything. I will put in a minor version release, and if we hear back from anyone after it is gone we can deal with it then.

PR #670, opened 2022-05-27
Store.groupby enhancements

### LINKED ISSUE #622: MongoLike groupby ignore sort, skip, limit kwargs
Thanks to clever synatx highlighting in my code editor, I noticed that the kwargs `sort`, `skip`, and `limit` that are shown for [`MongoStore.groupby`](https://github.com/materialsproject/maggma/blob/0c1ea7facbd97d53dc8df7b22dbbafc00f2eb92f/src/maggma/stores/mongolike.py#L255) and [`MemoryStore.groupby` ](https://github.com/materialsproject/maggma/blob/0c1ea7facbd97d53dc8df7b22dbbafc00f2eb92f/src/maggma/stores/mongolike.py#L615) appear to be non-functional.

These kwargs should be made functional (preferably, if it makes sense) or removed.

Related - it is not clear to me why the `groupby` methods for `MongoStore` and `MemoryStore` need to be different, since `MemoryStore` is designed to emulate `MongoStore` in every respect. If `MemoryStore` could inherit `MongoStore`'s `groupby` that would be more robust (and prevent subtle bugs like the one described in #621 ).

I may investigate this at a later time but would welcome comments from better-informed developers!

rkingsbury:
This PR began as an effort to fix #622 although has evolved based on what I learned. I've discovered a couple of inconsistencies related to `groupby` behavior and kwargs that I think we should rectify (see below). @munrojm what do you think is the best way to proceed here?

### behavior of `properties`

Apparently the default behavior in MongoDB is that when `properties` is `None`, `query` will return _all_ fields in a document, but `groupby` will return _nothing except the 'id' field_. See this from the [docs](https://www.mongodb.com/docs/manual/reference/operator/aggregation/project/)

> The _id field is, by default, included in the output documents. To include any other fields from the input documents in the output documents, you must explicitly specify the inclusion in [$project](https://www.mongodb.com/docs/manual/reference/operator/aggregation/project/#mongodb-pipeline-pipe.-project).

Personally I think within `maggma` it might make more sense to make the behavior consistent between `query` and `groupby` since not all `maggma` users will be intimately familiar with MongoDB conventions, but I'm not sure if we want to depart from what MongoDB does or not.

### `sort`, `skip`, and `limit` kwargs

A related point - `Store.groupby` defines the `sort`, `skip`, and `limit` kwargs, but it is not clear whether those make sense in the context of a groupby operation. They don't appear to be arguments in pymongo / MongoDB aggregation and they don't appear to be functional anywhere within `maggma`. So I would advocate that we either make those kwargs functional (and consistent with their behavior in `query`), or we remove them from the interface.

## Contributor Checklist

   -  [x] Clarify documentation of the `properties` kwarg in `query` and `groupby`
   -  [x] inherit `groupby` from `MongoStore` in `MemoryStore`
   - [x] Move `MemoryStore.groupby` to `MontyStore.groupby` b/c `montydb` does not implement an `aggregate` method and therefore can't inherit from `MongoStore.groupby`
   - [x] add tests of `sort`, `skip`, and `limit` kwargs in `MongoStore.query`
   - [x] Flag places in code where groupby kwargs `sort`, `skip`, and `limit` are non-functional


rkingsbury:
Hmmm, it looks like the test failure occurs because `montydb.collection` does not implement its own `aggregate` method, which is called by `MongoStore.groupby` whereas previously this was using the `MemoryStore.groupby` code. Strangely, this test passes for me locally with `montydb==2.3.12`.

rkingsbury:
> Hmmm, it looks like the test failure occurs because `montydb.collection` does not implement its own `aggregate` method, which is called by `MongoStore.groupby` whereas previously this was using the `MemoryStore.groupby` code. Strangely, this test passes for me locally with `montydb==2.3.12`.

OK, no idea why the test passes locally, but `montdyb` definitely does not implement `aggregate`: https://github.com/davidlatwe/montydb/search?q=aggregate

munrojm:
Hey @rkingsbury, thanks for this PR. 
'''

munrojm:
Hey @rkingsbury, thanks for this PR!

> Personally I think within maggma it might make more sense to make the behavior consistent between query and groupby since not all maggma users will be intimately familiar with MongoDB conventions, but I'm not sure if we want to depart from what MongoDB does or not.

I think I agree with this. Having the behavior consistent between the two appears to be less confusing to me as well.

> A related point - Store.groupby defines the sort, skip, and limit kwargs, but it is not clear whether those make sense in the context of a groupby operation. They don't appear to be arguments in pymongo / MongoDB aggregation and they don't appear to be functional anywhere within maggma. So I would advocate that we either make those kwargs functional (and consistent with their behavior in query), or we remove them from the interface.

I vote to make these functional here. I see your argument, but I feel as though including them will make using `groupby` possible in certain cases that might otherwise be difficult based on the total number of docs needing to be returned. Those query operations should be easy to add as the last set of stages in the aggregation pipeline `($sort -> $skip -> $limit)`. https://www.mongodb.com/docs/manual/reference/operator/aggregation-pipeline/.

PR #707, opened 2022-09-02
Add ssh_tunnel option to GridFSStore

utf:
Copied the same implementation as in MongoStore.

Would it be possible to get a new release with this functionality.


munrojm:
Absolutely! Thanks @utf .

PR #735, opened 2022-10-25
JSONStore: fix last_updated serialization problem

rkingsbury:
`datetime` objects serialize to `str` when using the `orjson` library, which breaks the `last_updated` functionality when a `JSONStore` is written and read back from disk. This PR fixes that problem by ensuring that the `last_updated` field, which is stored as a `str`, is properly deserialized into a `datetime`.

Also adds a new test and an additional test of the `last_updated` functionality in `MontyStore`.


munrojm:
Thanks @rkingsbury! Happy for me to merge?

rkingsbury:
Yep I'm good for you to merge!

PR #736, opened 2022-10-25
FileStore: fix metadata overwriting path

rkingsbury:
When a user adds metadata to ` FileStore` it is stored on disk. Both the `file_id` and the `path` are retained in the .json file to make it possible to manually associated metadata with specific files. However, when the metadata was read into the `FileStore`, the `path` attribute stored on disk was overwriting the `path` attribute determine by `FileStore.__init__`. This was not intended behavior, and caused problems when working with a `FileStore` across different computers where the absolute path of files is different (but the path relative to the `FileStore` root is the same). (For example, syncing a directory of files containing a `FileStore` across different computers using Dropbox, etc.)

This PR fixes the bug as follows:

- Store a new field `path_relative` instead of `path` in the metadata .json file on disk. This preserves manual identifiability of metdata
- Ensure that the `path` and `last_updated` fields cannot be overwritten by metadata updates
- add tests

### DIFF
--- a/src/maggma/stores/file_store.py
+++ b/src/maggma/stores/file_store.py
@@ -360,7 +360,11 @@ class FileStore(MemoryStore):
         Args:
             d: Dictionary whose keys are to be filtered
         """
-        filtered_d = {k: v for k, v in d.items() if k not in PROTECTED_KEYS}
+        filtered_d = {
+            k: v
+            for k, v in d.items()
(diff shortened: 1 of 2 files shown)


munrojm:
This looks good! Ready to merge?

rkingsbury:
> This looks good! Ready to merge?

Yes, ready!

PR #737, opened 2022-10-25
python 3.11 support + expand python versions tested in CI

rkingsbury:
- Add python 3.10 and 3.11 to CI test suite
- bump `numpy` to 1.23 to enable python 3.11 support
- bump `montdb` to 2.5.0 to enable python 3.11 support

Note: Per[ NEP 29,](https://numpy.org/neps/nep-0029-deprecation_policy.html) we are past the Python 3.7 support window, but since we are still supporting python 3.7 (it's not causing any problems); we should leave it in the CI test matrix.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -59,7 +59,7 @@ jobs:
       max-parallel: 6
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.7, 3.8, 3.9, 3.10, 3.11]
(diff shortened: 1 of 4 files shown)


rkingsbury:
Something going on with `3.10` related to https://github.com/actions/setup-python/issues/160. Will investigate later

rkingsbury:
Looks like [montydb](https://github.com/davidlatwe/montydb/tags) is a blocker for python 3.11 support at the moment.

PR #789, opened 2023-05-22
Add ruamel-yaml as a dependency

sivonxay:
Pyyaml is not a requirement of maggma, despite there being imports in `maggma.stores.mongolike` and `maggma.stores.gridfs`. I surmise this passes the tests because pyyaml is already installed when running the testing workflow.

While we're adding this dependency, we might as well start using ruamel-yaml, which other MP codes use.

### DIFF
--- a/requirements-docs.txt
+++ b/requirements-docs.txt
@@ -3,3 +3,4 @@ mkdocs-material==8.3.9
 mkdocs-minify-plugin==0.5.0
 mkdocstrings==0.18.1
 jinja2<3.2.0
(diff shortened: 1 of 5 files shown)


sivonxay:
I'm not sure why the docs are failing to build. They seem to be working locally for me.

sivonxay:
Aha. The docs don't install the requirements. I needed to add `ruamel.yaml` to the requirements-docs.txt file

munrojm:
No worries! Thank you for this fix as well.

PR #790, opened 2023-05-26
Implementation of an AzureBlobStore for Azure blobs

gpetretto:
This PR implements a Store for storing data in Azure blobs.  
The AzureBlobStore was modelled on the S3Store with similar logic and similar options (an index Store for the base data and queries, with references to the blob ids).
The two basic authentication methods described in the [azure tutorial](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-quickstart-blobs-python?tabs=managed-identity%2Croles-azure-portal%2Csign-in-azure-cli) have been implemented, but left open for potential future additional methods. (Note: the suggested authentication method changed not long ago).
Tests have been implemented using [Azurite](https://github.com/Azure/Azurite)

## Contributor Checklist


munrojm:
@gpetretto this looks great! Thanks for all the work put into it. I think it will be a great addition. 

No issues on my end since it is modeled so closely after the S3Store. I may try and factor out a more generic BlobStore class from the two to avoid some of the repetition, but would like to merge this is now and minor release if you are good with that.

gpetretto:
Hi @munrojm, indeed I have also thought that it should be possible to extract all the basic functionalities between the S3Store and AzureBlobStore. This would also simplify the implementation of similar Stores relying on other cloud services. However, I wanted to have a working version for Azure first for a project we are running. So it would be good to merge now.

Issue #733, opened 2022-10-24
GridFSStore collection has no database attribute


PanayotisManganaris:
Any atomate2 TaskDocument larger than the document limit cannot spill into gridfs additional stores successfully because the gridfs collection has no database property

https://github.com/materialsproject/maggma/blob/0680fa9508fe29e8e84963d54662a6c9e3e94d55/src/maggma/stores/gridfs.py#L450

versions:
`db> db.version() `
6.0.2

I deployed the official docker container on an enterprise server.

`conda list` (truncated)
`jobflow                   0.1.7              pyhd8ed1ab_0    conda-forge`
`maggma                    0.49.4             pyhd8ed1ab_0    conda-forge`
`atomate2                  0.0.1.post912+g1fd2322          pypi_0    pypi` (my fork main)
`pymongo                   4.2.0                    pypi_0    pypi`

Very new to all this. Advice appreciated.

munrojm:
Hi @PanayotisManganaris. I will take a look at this.

PR #791, opened 2023-05-29
Add orjson options in JSONStore

gpetretto:
orjson, used for serialization in JSONStore, supports different options for serialization. This PR adds the possibility to pass down the `option` and `default` keywords for serialization, that allow to tune orjson's serialization behaviour.
In my case, I need to serialize the output of an atomate calculation and it contains FloatWithUnit, which is subclass of float and as such is not handled by orjson (https://github.com/ijl/orjson/issues/114).

I have also fixed the `close` method of the `GridFSStore`, that was raising an exception.

## Contributor Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,7 +15,7 @@ uvicorn==0.18.3
 sshtunnel==0.4.0
 msgpack==1.0.3
 msgpack-python==0.5.6
-orjson==3.8.0
+orjson==3.8.14
 boto3==1.24.42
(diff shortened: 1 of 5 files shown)


gpetretto:
I needed to bump orjson version, since the options were not available in the older version.

I have also noted that the fix in the GridFSStore should solve https://github.com/materialsproject/maggma/issues/733

munrojm:
Looks great! Thank you for this as well.

PR #794, opened 2023-06-07
Memray memory profiler support for mrun command line tool

tsmathis:
This PR adds an additional optional flag to `mrun`: `--memray`, or alternatively `-m`, to enable memory profiling of a builder using the Memray Python memory profiling tool ([Memray repo](https://github.com/bloomberg/memray)).

The profiler supports profiling of both single and forked processes. For example, spawning multiple processes in `mrun` with `-n` will signal the profiler to track any forked child process spawned from the parent process.

A basic invocation from a terminal would look something like `mrun --memray on builder_dump_file.json`. `--memray` is a Click boolean parameter ([Click](https://click.palletsprojects.com/en/5.x/parameters/)) and thus accepts the values `1`, `yes`, `y`, `on`, and `true`, which convert to `True`, and `0`, `no`, `n`, `off`, and `false`, which convert to `False`.

The output `.bin` file produced by Memray is dumped by default into the user's `temp` directory, which is platform/OS dependent. For Linux/MacOS this will be `/tmp/` and for Windows the target directory will be `C:\TEMP\`.

The output file will have a generic naming pattern as follows: `BUILDER_NAME_PASSED_TO_MRUN + BUILDER_START_DATETIME_ISO.bin`, e.g., `builder_dump.json_2023-06-06T16.14.19.272030.bin`.

Following completion, the `.bin` output file can be converted into a flamegraph using the Memray CLI, i.e, `memray flamegraph builder_dump.bin`.

Further visualization and data transform examples can be found in Memray's documentation ([Memray reporters](https://bloomberg.github.io/memray/live.html)).

### DIFF
--- a/docs/getting_started/running_builders.md
+++ b/docs/getting_started/running_builders.md
@@ -119,3 +119,23 @@ mrun -n 32 -vv my_first_builder.json builder_2_and_3.py last_builder.ipynb
 * `BUILD_ENDED` - This event tells us the build process finished this specific builder. It also indicates the total number of `errors` and `warnings` that were caught during the process.
 
 These event docs also contain the `builder`, a `build_id` which is unique for each time a builder is run and anonymous but unique ID for the machine the builder was run on.
+
+
+## Profiling Memory Usage of Builders
+
+`mrun` can optionally profile the memory usage of a running builder by using the Memray Python memory profiling tool ([Memray](https://github.com/bloomberg/memray)). To get started, `maggma` will first need to be installed from source ([Maggma installation](https://materialsproject.github.io/maggma/#installation-from-source)) followed by `pip` installing Memray using `pip install memray`, or by installing the optional `maggma` requirements by using `pip install requirements-optional.txt` in the `maggma` base directory.
+
+Setting the `--memray` (`-m`) option to `on`, or `True`, will signal `mrun` to profile the memory usage of any builders passed to `mrun` as the builders are running. The profiler also supports profiling of both single and forked processes. For example, spawning multiple processes in `mrun` with `-n` will signal the profiler to track any forked child processes spawned from the parent process.
+
+A basic invocation of the memory profiler using the `mrun` command line tool would look like this:
+``` shell
+mrun --memray on my_builder.json
+```
+
+The profiler will generate two files after the builder finishes: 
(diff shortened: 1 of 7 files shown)


rkingsbury:
This looks useful, thanks @tsmathis! I just have a few comments:

1. To keep `maggma` as light as possible, I think `memray` should be included in `requirements-optional.txt` rather than core `requirements.txt`, and excluded from `setup.py`. @munrojm let us know if you have other thoughts
2. Please add a simple test. While I don't think we need to add a test of the `memray` output per se, we should have a test that verifies `mrun` works as expected when the `memray` flag is invoked, and also that specifying a non-default directory for the memory profile works
3. Please add a section to the Running Builders docs explaining how this option works. The contents of your first post on this PR would be just about perfect

tsmathis:
Thanks for the comments, @rkingsbury! I will address your points before reopening!

munrojm:
I second what @rkingsbury has suggested. Other than that I’m good with merging.

tsmathis:
1. An additional flag ,`-md`, has been added to allow specification of a non-default output directory.
2. Two tests added to `test_init.py` verifying that (1) `mrun` behaves as expected when `memray` is enabled, and (2) specifying a non-default directory works.
3. A section ("Profiling Memory Usage of Builders") describing usage of the `memray` flag with `mrun` and its options has been added to the running builders docs

docs/getting_started/running_builders.md
```diff
@@ -119,3 +119,25 @@ mrun -n 32 -vv my_first_builder.json builder_2_and_3.py last_builder.ipynb
 * `BUILD_ENDED` - This event tells us the build process finished this specific builder. It also indicates the total number of `errors` and `warnings` that were caught during the process.
 
 These event docs also contain the `builder`, a `build_id` which is unique for each time a builder is run and anonymous but unique ID for the machine the builder was run on.
+
+
+## Profiling Memory Usage of Builders
+
+`mrun` can optionally profile the memory usage of a running builder by using the Memray Python memory profiling tool ([Memray](https://github.com/bloomberg/memray)). To get started, `maggma` will first need to be installed from source ([Maggma installation](https://materialsproject.github.io/maggma/#installation-from-source)) followed by `pip` installing Memray using `pip install memray`, or by installing the optional `maggma` requirements by using `pip install requirements-optional.txt` in the `maggma` base directory.
```
rkingsbury, line 126: Why is it necessary to install `maggma` from source in order for this to work? Is it not possible to `pip install maggma` and then install the optional `memray`?

rkingsbury:
Thanks for addressing comments @tsmathis , this looks great! I just have one small question about the installation requirements; otherwise I think this is ready to merge.

docs/getting_started/running_builders.md
tsmathis: Ah you are right, I will correct that

tsmathis: Fixed!

rkingsbury:
Thanks @tsmathis !

PR #795, opened 2023-06-08
Use tqdm.auto

utf:
Use `tqdm.auto` to automatically detect whether you are in a notebook and avoid warnings on tqdm imports.


utf:
Other changes were made by the pre-commmit.

rkingsbury:
Thanks for this nice enhancement @utf ; I was not aware of `tqdm.auto` before but now I'm sure I'll use it all the time!

Issue #807, opened 2023-06-14
python 3.11 CI test failure with AzureBlobStore


rkingsbury:
Since adding python 3.11 support (See #737 ), `tests.stores.azure.test_multi_update` frequently (but not always) fails using python 3.11. See this [example workflow run](https://github.com/materialsproject/maggma/actions/runs/5260085303/jobs/9506538272#step:6:123).

I don't know the exact cause, but the issue is that the test condition is based on the execution time, with the _assumption_ that a multiprocessed write will be faster than a single write:

```
        start = time.time()
        blobstore_multi.update(data, key=["task_id"])
        end = time.time()
        time_multi = end - start
    
        start = time.time()
        blobstore_two_docs.update(data, key=["task_id"])
        end = time.time()
        time_single = end - start
>       assert time_single > time_multi * (blobstore_multi.workers - 1) / (
            blobstore_two_docs.workers
        )
E       assert 0.26502466201782227 > ((0.5272753238677979 * (4 - 1)) / 4)
E        +  where 4 = <maggma.stores.azure.AzureBlobStore object at 0x7f0c5daa4810>.workers
E        +  and   4 = <maggma.stores.azure.AzureBlobStore object at 0x7f0c5daa4810>.workers
```

For whatever reason, in python 3.11 it seems that at least sometimes the single write is faster. I'm not sure whether this is due to a change from python 3.10 to 3.11, or some issue with GitHub python 3.11 workers, but regardless, I wonder @gpetretto if there is a different way you could write the `assert` statements that would test the functionality without being dependent on performance?

PR #808, opened 2023-06-14
Fixing bug in azure multi worker test

gpetretto:
Following #807 I have realized that there was a bug in the referred test (a leftover of importing the tests from those of the `S3Store`). Hopefully this quick fix may already solve the issue. I cannot verify locally, since I have tried to run the test on my laptop with a python 3.11 environment and it always passed, even with the bugged version. 

If this will not be enough I will try to devise a different kind of check.


rkingsbury:
Thanks for the quick action on this @gpetretto !

gpetretto:
Thank you for reaching out, otherwise I probably would have not noticed. It seems it could have solved the issue. Ping me again in case it starts failing in some other PR.

PR #809, opened 2023-06-22
add patch method for submission resource

yang-ruoxi:
* Added `build_patch_data` in Submission resource to update fields in a collection 
* The  will accept the `query['update']` parameters to update corresponding fields in a collection, using the `$set` operator: 
    `self.store_collection.update_one(filter=query['criteria'], {'$set': query["update"]})`
* To accommodate this change, added `"update"` into `STORE_PARAMS`


yang-ruoxi:
thanks, and good point! Will do that

PR #829, opened 2023-07-25
allow >5GB and turn on multi-part uploads for AWS

kbuma:
*Change AWS upload code to allow >5GB objects to be saved and to turn on multi-part uploads*

## Contributor Checklist

- [X] I have run the tests locally and they passed.


munrojm:
@kbuma thanks again for taking a look at this and making the necessary changes.

munrojm:
Do you have any concerns with the upstream multithreading of that function? In other words, does it even make sense to keep the threadpool upstream, or should be just let boto3 handle concurrency via multithreaded chunk uploads of single blobs in succession?

kbuma:
from my understanding the upstream multithreading is dealing with parallelizing multiple file uploads. this update will parallelize the multiple parts of a single file upload. so they could be complimentary.
the updated code improves performance for large files whereas the upstream multithreading could provide you performance enhancements in situations where you have a lot of small files.

PR #837, opened 2023-08-01
Allow maggma to be used without Azure

jmmshn:
Currently, 0.52.0 raises an error

```
../../../local/Caskroom/miniconda/base/envs/mp/lib/python3.9/site-packages/maggma/stores/azure.py:368: in AzureBlobStore
    def _get_container(self) -> Optional[ContainerClient]:
E   NameError: name 'ContainerClient' is not defined
```


munrojm:
Do you have a sense of why this didn't happen before? Is it from dep upgrades?

jmmshn:
A line was commented out here:
https://github.com/materialsproject/maggma/commit/6690d0a822e052470942c475491f69744090b951

PR #841, opened 2023-08-07
CI: use OS-specific requirements in testing

rkingsbury:
Use OS-specific `requirements` files generated automatically by the GitHub workflow when setting up CI environments.


arosen93:
You can probably drop Python 3.7 tests altogether since it has reached its end-of-life.

PR #843, opened 2023-08-15
Aws store botocore fix

tsmathis:
## Summary
Major changes:
This PR fixes a bug in the botocore ClientError response checking that occurs when an S3 store is queried. Previously, checks were made for a `404` client error code, but S3 only returns string error codes, e.g., `"NoSuchKey"`. 

The proposed changes instead check for the error codes that might be commonly encountered during the running of a builder (`NoSuchKey` & `NoSuchBucket`), logs instances of these errors, and then allows the builder to continue. Other errors continue to raise an exception, consistent with previous behavior.

The test associated with checking the `404` error code have been updated accordingly.

- [x] Tests updated for new features/fixes.

### DIFF
--- a/src/maggma/stores/aws.py
+++ b/src/maggma/stores/aws.py
@@ -187,7 +187,7 @@ class S3Store(Store):
                     error_code = e.response["Error"]["Code"]
                     if error_code == "NoSuchKey":
                         self.logger.error(f"Could not find S3 object {doc[self.key]}")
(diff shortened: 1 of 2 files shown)


tsmathis:
@munrojm, here are those changes to the aws store ClientError checks

munrojm:
Great, the changes look good. I will go ahead and merge.

PR #847, opened 2023-08-28
JSONStore: enabled reading of MongoDB extended JSON files

rkingsbury:
## Summary

When exporting a MongoDB collection to JSON, the resulting format is [extended JSON](https://www.mongodb.com/docs/manual/reference/mongodb-extended-json/#bson.ObjectId). Trying to load such a JSON file into `JSONStore` currently fails due to the way extended JSON serializes `ObjectID`, with an error similar to

```
bson.errors.InvalidDocument: key '$oid' must not start with '$'
```

This PR modifies `JSONStore.read_json` to use the [`bson.json_util.loads`](https://pymongo.readthedocs.io/en/stable/api/bson/json_util.html#bson.json_util.loads) method if `$oid` is detected in the file, which allows loading extended JSON files into a `JSONStore`. If `$oid` is not there, the current `orjson.loads()` is used instead.

Unfortunately. `orjson.loads` does not support BSON types. This was [requested before](https://github.com/ijl/orjson/issues/70), although it isn't clear if the maintainer understood the request.

Happy to hear opinions if there are better ways to accomplish this, e.g. by leveraging `serialization_helper`, `MontyDecoder`, etc. and/or opening another `orjson` issue with a more specific request about how to handle these types.

### DIFF
--- a/src/maggma/stores/mongolike.py
+++ b/src/maggma/stores/mongolike.py
@@ -18,6 +18,7 @@ except ImportError:
 
     from typing_extensions import Literal
 
+import bson
 import mongomock
 import orjson
 from monty.dev import requires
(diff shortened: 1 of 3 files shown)


rkingsbury:
Flagging @arosen93 as well for any comments

arosen93:
No comments from me here. Looks great!

rkingsbury:
@munrojm , any thoughts before I merge?

munrojm:
@rkingsbury Nope! Looks good to me.

munrojm (approved):
LGTM!

PR #850, opened 2023-08-30
Update CI pipeline so repeated commits don't cause concurrent tests

arosen93:
I updated the CI pipeline such that if multiple commits have been made in relatively quick succession, the prior commit's in progress tests will be cancelled. Since the maggma tests take a fair bit of time, this is a nice convenience feature in my opinion.

I also updated the classifier for the production status to "Development Status :: 5 - Production/Stable" and added a missing `src` line in the `ruff` spec for `pyproject.toml`.

Note: I first learned about this from `repo-review` and have been using it for all my projects. You might be interested in some of [the items flagged for maggma](https://scientific-python.github.io/repo-review/?repo=materialsproject%2Fmaggma&branch=main), although a fair number of them are quite subjective.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -8,6 +8,10 @@ on:
   schedule:
     - cron: "11 19 * * 6"
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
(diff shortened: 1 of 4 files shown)


rkingsbury:
Good stuff, thank you @arosen93 ! I will be making use of repo-review 😄

PR #851, opened 2023-08-30
Remove tests for Python 3.7 since it reached its end-of-life

arosen93:
Python 3.7 reached its [end-of-life](https://devguide.python.org/versions/) on 06-27-2023. As such, I removed it from the test matrix and from the classifiers list in `setup.py`. I didn't change the `python_requires` minimum version in `setup.py` though to allow for some flexibility for those that are slow to upgrade.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -47,7 +47,7 @@ jobs:
       max-parallel: 6
       matrix:
         os: [ubuntu-latest]
-        python-version: ["3.7","3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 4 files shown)


rkingsbury:
Thanks @arosen93 ! This has been on my mind for a while but I wasn't sure when to pull the trigger. You have convinced me. Can you please also update the docs to "3.8+"? There is this line:

> Maggma is written in [Python](http://docs.python-guide.org/en/latest/) and supports Python 3.7+.

> I didn't change the python_requires minimum version in setup.py though to allow for some flexibility for those that are slow to upgrade.

Tell me more about this. I suppose this means that a user _could_ install `maggma` in a Python 3.7 environment, whereas if we change it to 3.8, they would be prevented from doing so? I'm on the fence about this approach; I kinda feel like as long as it is possible to install with Py3.7, we ought to be testing against it. But it is becoming a pain to support so I'm in favor of dropping it.

rkingsbury:
Side note, this question of how long to support 3.7 was a big part of what motivated my [foundation PR on version support](https://github.com/materialsproject/foundation/pull/11)

`emmet` and `atomate2` already set `python_requires=3.8`, and since those are probably the downstream packages that use `maggma` most, I think we're safe to drop 3.7

arosen93:
@rkingsbury: I agree. I think consistency is smart and there shouldn't be any issues here. I have now updated the PR to reflect this discussion, including both the update to the docs and the `setup.py`.

rkingsbury:
Given that pymatgen is now at 3.9+, I'm gonna pull the trigger on this. 

See https://github.com/materialsproject/pymatgen/pull/3283

PR #852, opened 2023-08-30
Create dependabot.yml to update GitHub actions packages

arosen93:
See https://learn.scientific-python.org/development/guides/gha-basic/#GH200 for context.


arosen93:
Ah okay! Good catch! I was wondering why there wasn't a depdendabot file originally. That explains it!

arosen93:
@rkingsbury: Should be good to go now.

rkingsbury:
Thanks!

PR #882, opened 2023-11-07
SSH tunnel support for S3Store

mjwen:
## Summary

This PR adds the support to use an ssh tunnel for `S3Store`, which can be useful for self-hosted MinIO. This follows the same pattern as the ssh tunnel support for `MongoStore`. 

- Created a new file `ssh_tunnel.py` and moved the SSHTunnel class and related stuff there. It's a bit weird to put it in the monogostore file. 
- Added a new optional parameter `local_port` for the SSHTunnel class. This is necessary for MinIO, by default which uses the  9000 port. The original implementation that uses an automatically chosen free port will not work in general. 
- Added `ssh_tunnel` support for the `S3Store`.  
- Added a tutorial doc to use `SSHTunnel`.


rkingsbury:
Thanks @mjwen , this looks like a nice enhancement. After a quick review, can I ask for two changes?

1. Can you please expand the test coverage of SSHTunnel? E.g. add tests that attributes get set correctly, etc.? I'd really like to see test coverage for the patch close to 100%; and in any case I don't want the project test coverage to decrease.
2. In `S3Store.connect`, please make sure to implement `force_reset` in a way that is consistent with other stores (see #879 ; comments welcome on how that works!)

Finally (and this is a bigger ask), if you'd be able to add a page to the docs describing how to connect via SSH, I'm sure many users would appreciate it. But I understand if that's not possible at this time.

mjwen:
@rkingsbury thanks for the quick feedback! 

Great that the functionality looks alright to you. I will add the enhancements (tests and docs) you mentioned above and ping you once ready.

mjwen:
@rkingsbury I've addressed your comments. A couple of stuff:

1. Seems the test GH action is not set up to enable database server, e.g. localhost:22 cannot be access. As a results, many of the tests (old and the new ones I've added) involving database are skipped. Then the coverage tests cannot cover them as well. We should somehow enable this? I've no idea how to do it in GH action. @jmmshn might have some experience? 

2. I did not find anything on automatic docs build and publish in the [docs](https://github.com/materialsproject/maggma/tree/main/docs) directory. Is it located somewhere else? The [current docs](https://materialsproject.github.io/maggma/) seem old -- some of the md pages are now included, e.g. the docs on [FileStore](https://github.com/materialsproject/maggma/blob/main/docs/getting_started/using_file_store.md) you've added.

jmmshn:
Is there some way to spoof the ssh_tunnel connection?
I think `paramiko` should not actually be called during any of the CI just like how we are using moto to mock all the bots connections.

So I think if the basic tests are designed only in a way to check the arguments to `paramiko` then we are OK.

For more complex tests, you can consider mocking the entry functions to the ssh_tunnel and have that turn some kind of a flag, then you just have to check that connecting to a DB properly hits that flag?
Basically, I think our CI should only check that settings are properly passed to SSH tunnel API but we don't need to actually check the SSH tunnels.

rkingsbury:
Thanks @mjwen !

>     1. Seems the test GH action is not set up to enable database server, e.g. localhost:22 cannot be access. As a results, many of the tests (old and the new ones I've added) involving database are skipped. Then the coverage tests cannot cover them as well. We should somehow enable this? I've no idea how to do it in GH action. @jmmshn might have some experience?

I'll defer to Jimmy on this one, and I think the approach he suggested sounds ok (just make sure we are passing the right arguments, not testing the tunnel itself)

>     2. I did not find anything on automatic docs build and publish in the [docs](https://github.com/materialsproject/maggma/tree/main/docs) directory. Is it located somewhere else? The [current docs](https://materialsproject.github.io/maggma/) seem old -- some of the md pages are now included, e.g. the docs on [FileStore](https://github.com/materialsproject/maggma/blob/main/docs/getting_started/using_file_store.md) you've added.

I'm not sure I follow - as far as I can tell, the current docs you linked are up to date. I think you just need to add modify a .`md` file in `docs/reference` (or wherever you think).

I guess the docs build process isn't document, but we use `mkdocs`. So if you want to preview your changes, you can run `mkdocs serve` and then it should serve your documentation in the browser, and refresh in real time as you edit. Does that help?

mjwen:
@rkingsbury My bad regarding the docs! I was looking for some configure file in the `docs` directory, but now I see `mkdocs.yml` is in the project root directory. Updated.

mjwen:
@jmmshn Thanks for the suggestions! Sounds like a good compromise for the CI tests. I've added some fake tunnel to test the functionality s3store.

mjwen:
@rkingsbury Ready for final review. 

Note, the CI won't test some of the attributes assignments of the SSHTunnel because of the way we are setting it up. 

Also, the CI tests seem fail. I've taken a quick look at it -- seems related to `pyproject.toml` and `setup.py`. A rerun might work?

tests/stores/test_aws.py
```diff
@@ -125,7 +151,7 @@ def test_count(s3store):
     assert s3store.count({"task_id": "mp-3"}) == 1
 
 
-def test_qeuery(s3store):
+def test_qeury(s3store):
```
rkingsbury, line 154: typo `test_qeury` -> `test_query`

docs/getting_started/using_ssh_tunnel.md
```diff
@@ -0,0 +1,56 @@
...
+    username = "<USERNAME>",
+    password= "<USER_CREDENTIAL>",
+    remote_server_address = "COMPUTE_NODE_1:9000", 
+    local_port = 9000,
+)
+```
+and then pass it to the `S3Store` to connect to the database. 
+By doing so, you can access the database at the localhost address `http://127.0.0.1:9000` from your local computer as if it is hosted on your local computer. 
+
+The arguments of the `SSHTunnel` are self-explanatory, but `local_port` needs more explanation. We assume that on the local computer, we want t o connect to the localhost address `http://127.0.0.1`, so we do not need to provide the address, but only the port number (`9000` in this case.)
```
rkingsbury, line 54: small typo `t o` -> `to`

docs/getting_started/using_ssh_tunnel.md
```diff
@@ -0,0 +1,56 @@
...
+
+tunnel = SSHTunnel(
+    tunnel_server_address = "<REMOTE_SERVER_ADDRESS>:22",
+    username = "<USERNAME>",
+    password= "<USER_CREDENTIAL>",
+    remote_server_address = "COMPUTE_NODE_1:9000", 
+    local_port = 9000,
+)
+```
+and then pass it to the `S3Store` to connect to the database.
```
rkingsbury, line 51: Would you mind mentioning that other store classes can use this to? On a quick review of the codebase, I think `GridFSStore`, `MongoStore`, and MongoURIStore` all support ssh_tunnel.

tests/stores/test_ssh_tunnel.py
```diff
@@ -21,10 +19,24 @@
         pytest.skip("No SSH server to test tunnel against")
 
 
-def test_mongostore_connect_via_ssh(ssh_server_available):
-    server = SSHTunnel("127.0.0.1:22", "127.0.0.1:27017")
+def local_port_available(local_port):  # noqa: PT004
+    """Fixture to determine if a local port is available to test the SSH tunnel."""
+    client = paramiko.SSHClient()
+    client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
```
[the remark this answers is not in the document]
rkingsbury: Would the tests work if you were to set `client.set_missing_host_key_policy(RejectPolicy)` ? If so, please do that. If not, it's OK b/c this is just a test file so I'm not so concerned about security issues like this.

See here for more details: https://github.com/materialsproject/maggma/security/code-scanning/138

rkingsbury (changes requested):
Thanks @mjwen , looks good. I just have a few minor typo edits to suggest.

Apologies for the CI issues; apparently there was a setuptools update or something that broke the tests. They should be working now if you pull the latest.

tests/stores/test_aws.py
mjwen: Done

docs/getting_started/using_ssh_tunnel.md
mjwen: Done

docs/getting_started/using_ssh_tunnel.md
mjwen: Done

mjwen:
I've checked `client.set_missing_host_key_policy` a bit, and it seems `client.set_missing_host_key_policy(RejectPolicy())` is the default and it won't work with the tests, because we need to load the host key info for the tests. 

I believe we need to leave it like that.

rkingsbury:
Thanks @mjwen !

PR #893, opened 2023-11-22
Enhancement/open data store

kbuma:
## Summary

Major changes:

- add S3IndexStore: a store that loads the index of the collection from an S3 file.
- add OpenDataStore: a store whose data is stored on S3 compatible storage using the format used by Materials Project on OpenData. It utilizes an index that is loaded from S3 compatible storage into memory (S3IndexStore).
- refactor of S3Store to make it extensible, ie allow OpenDataStore to build on top of it

## Checklist

### DIFF
--- a/src/maggma/stores/aws.py
+++ b/src/maggma/stores/aws.py
@@ -184,7 +184,7 @@ class S3Store(Store):
             else:
                 try:
                     # TODO: This is ugly and unsafe, do some real checking before pulling data
-                    data = self.s3_bucket.Object(self.sub_dir + str(doc[self.key])).get()["Body"].read()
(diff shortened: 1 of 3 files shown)


munrojm:
@kbuma just went through this in detail. Everything is looking fantastic. I like the structure, and don't really have any specific comments other than to get your input on making the type of compression more configurable for the user. I'm thinking there may be some value in having the compression/decompression callables and header passed in the __init__, but with the same gzip defaults. Don't have a strong opinion though, so happy to merge whenever you feel good.

kbuma:
@munrojm I'd like to hold off on making the compression more configurable unless there's an existing use case for it. If there is one let's discuss and add the feature. Otherwise, I'm good to merge.

PR #900, opened 2023-12-19
Patch s3 kwarg

jmmshn:
## Allow the Indexstore in S3 to be updated through Kwargs

Some programs like atomate1 uses the Kwargs for s3 during the configuration for the store.
As such there is no easy way to pass lesser used kwargs like ssl setting to the index store.
This gives a option for user to do that.

Also removed some of re-definitions of kwargs in the open data object that should just be piped to the S3Store parent class that was caught by codeQL


rkingsbury:
Thanks @jmmshn !

PR #908, opened 2024-01-18
creating PandasMemoryStore for use by OpenDataStore

kbuma:
## Summary

In testing we discovered that `MemoryStore` was not scaling and we were hitting performance walls once we got to 3-4K documents. For our use case we need a memory store that is performant with 500K+ documents. 

- create `PandasMemoryStore` that utilized DataFrames as backing store
- utilize this new store for `S3IndexStore` (and thus `OpenDataStore`) instead of `MemoryStore`

**NOTE:** This new store is not 100% compatible with code written for existing stores. Namely, the criteria dictionaries used for querying need to be rewritten such that the queries conform to Pandas query expressions (https://pandas.pydata.org/docs/user_guide/indexing.html#the-query-method) rather than Mongo style.

**NOTE:** To utilize this new store and the OpenData related stores you will need to install the `open_data` extra.

## Checklist

### DIFF
--- a/src/maggma/stores/open_data.py
+++ b/src/maggma/stores/open_data.py
@@ -1,17 +1,17 @@
 import gzip
+from datetime import datetime
 from io import BytesIO
 from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, Union
 
 import orjson
+import pandas as pd
 from boto3 import Session
 from botocore import UNSIGNED
 from botocore.config import Config
 from botocore.exceptions import ClientError
 from bson import json_util
-from datetime import datetime
-import pandas as pd
-from maggma.core.store import Sort, Store
 
+from maggma.core.store import Sort, Store
 from maggma.stores.aws import S3Store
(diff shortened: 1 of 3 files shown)


src/maggma/stores/open_data.py
```diff
@@ -1,19 +1,324 @@
...
+        skip: int = 0,
+        limit: int = 0,
+    ) -> Iterator[Tuple[Dict, List[Dict]]]:
+        """
+        Simple grouping function that will group documents
+        by keys.
+
+        Args:
+            keys: fields to group documents
+            criteria: PyMongo filter for documents to search in
```
tsmathis, line 278: `criteria` here shouldn't be a PyMongo filter, right? Should be same doc string as above?
`criteria: the value of the `query` key will be used as the Pandas string expression to query with;`

kbuma: fix added - thanks for catching this

PR #923, opened 2024-02-18
Tweak docstrings to fix rendered docs

ml-evs:
In a similar vein to https://github.com/materialsproject/api/pull/863, this PR makes several docstring fixes that are otherwise breaking the rendered mkdocs, making it quite hard for users to use the API without digging into the code.

- [x] Enable D411 ruff rule in pre-commit to ensure that sections are properly demarcated
- [x] Fix D415 missing punctuation at end of summary so that docstring formats get properly detected
- [x] Manually fix some broken docstrings in cases where indented code was erroneously treated as code snippets

Example from deployed docs:

2024-02-18-121757_705x632_scrot

Which will now look like:

2024-02-18-121734_781x657_scrot

There are still hundreds of pydocstyle warnings but I think this at least fixes the ones that mkdocs(trings) will care about.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,7 +13,7 @@ repos:
     rev: v0.0.280
     hooks:
       - id: ruff
-        args: [--fix, --show-fixes, --ignore, D]
+        args: [--fix, --show-fixes, --ignore, D, --extend-select, D411]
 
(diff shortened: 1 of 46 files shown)


rkingsbury:
Thanks very much @ml-evs !

PR #942, opened 2024-04-15
Add support for ruamel.yaml 0.18+ by replacing deprecated `.safe_load()` function

### LINKED ISSUE #938: [Feature Request]: support ruamel.yaml 0.18+
### Problem

https://github.com/materialsproject/maggma/blob/483962fb4ffed6568c33a56ce05f0cb64b180cd4/setup.py#L27

There is a downward version pin on ruamel.yaml that we should, eventually, address within maggma. 

### Proposed Solution

TBD. 

### Alternatives

_No response_

Andrew-S-Rosen:
## Summary

Closes #938.

This PR ensures maggma works with ruamel.yaml >= 0.18.0+ where the `.safe_load()` method was removed. The functionality is retained; the function is just called differently. The method was deprecated in 0.17.0 (2021-03-06) and removed in 0.18.0 (2023-10-23), so we have had ample time. 😅 

I think this should be good to go, @rkingsbury!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ setup(
     include_package_data=True,
     install_requires=[
         "setuptools",
-        "ruamel.yaml<0.18",
+        "ruamel.yaml>=0.18",
         "pydantic>=2.0",
         "pydantic-settings>=2.0.3",
         "pymongo>=4.2.0",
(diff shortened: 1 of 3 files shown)


rkingsbury:
Thanks for this, @Andrew-S-Rosen !

Issue #932, opened 2024-03-25
[Feature Request]: pass keyword arguments to zopen to accommodate non english platforms


rkingsbury:
### Problem

User needs to be able to pass kwargs through to `zopen` (e.g., in `JSONStore`) in order to accommodate non-english platforms. 

See https://github.com/materialsvirtuallab/monty/issues/641#event-12234573501

zopen fails on some language platforms due to lack of explicit encoding (UnicodeDecodeError) #641

### Proposed Solution

- Identify anywhere in `maggma` that uses `zopen`
- Add passthrough kwargs from the `Store.__init__` method to `zopen` so that users of other platforms can explicitly set character encoding

rkingsbury:
See also https://github.com/KingsburyLab/pyEQL/issues/122#issuecomment-2010944950

rkingsbury:
Closed by #949

PR #949, opened 2024-05-13
Add character encoding kwarg to JSONStore and FileStore

rkingsbury:
## Summary

Addresses #932 by adding a `encoding` kwarg to the `Store` that make use of `zopen`, i.e., `JSONStore` and `FileStore`. This should make it easier to troubleshoot cases when the automatically detected character encoding is not consistent with the files being read.


rkingsbury:
Closes #932

Issue #928, opened 2024-02-27
[Feature Request]: Leverage optional dependency groups to reduce dependency count


rkingsbury:
### Problem

As recently surfaced in a [peer review](https://github.com/openjournals/joss-reviews/issues/6295) of a package that depends on `maggma`, `maggma` introduces a LOT of dependencies (ultimately because it so versatile). However, any given user of `maggma` is probably only using it to connect 2 or 3 types of `Store` and might not make use of many of these.

```
├── maggma [required: >=0.57.5, installed: 0.63.3]
│   ├── aioitertools [required: >=0.5.1, installed: 0.11.0]
│   │   └── typing-extensions [required: >=4.0, installed: 4.10.0]
│   ├── boto3 [required: >=1.20.41, installed: 1.34.49]
│   │   ├── botocore [required: >=1.34.49,<1.35.0, installed: 1.34.49]
│   │   │   ├── jmespath [required: >=0.7.1,<2.0.0, installed: 1.0.1]
│   │   │   ├── python-dateutil [required: >=2.1,<3.0.0, installed: 2.8.2]
│   │   │   │   └── six [required: >=1.5, installed: 1.16.0]
│   │   │   └── urllib3 [required: >=1.25.4,<1.27, installed: 1.26.18]
│   │   ├── jmespath [required: >=0.7.1,<2.0.0, installed: 1.0.1]
│   │   └── s3transfer [required: >=0.10.0,<0.11.0, installed: 0.10.0]
│   │       └── botocore [required: >=1.33.2,<2.0a.0, installed: 1.34.49]
│   │           ├── jmespath [required: >=0.7.1,<2.0.0, installed: 1.0.1]
│   │           ├── python-dateutil [required: >=2.1,<3.0.0, installed: 2.8.2]
│   │           │   └── six [required: >=1.5, installed: 1.16.0]
│   │           └── urllib3 [required: >=1.25.4,<1.27, installed: 1.26.18]
│   ├── dnspython [required: >=1.16.0, installed: 2.6.1]
│   ├── fastapi [required: >=0.42.0, installed: 0.110.0]
│   │   ├── pydantic [required: >=1.7.4,<3.0.0,!=2.1.0,!=2.0.1,!=2.0.0,!=1.8.1,!=1.8, installed: 2.6.2]
│   │   │   ├── annotated-types [required: >=0.4.0, installed: 0.6.0]
│   │   │   ├── pydantic-core [required: ==2.16.3, installed: 2.16.3]
[...]

PR #950, opened 2024-05-14
MongograntStore: make optional dependency, prepare for removal

rkingsbury:
## Summary

As discussed in #928 

- Move `mongogrant` to the optional dependency group `[mongogrant]`
- Add a `requires` decorator to `MongograntStore`
- Mark `MongograntStore` for deprecation
- Remove from docs


rkingsbury:
Does this look OK to you @munrojm ? I agree with your comment that we should probably remove `MongograntStore` eventually. This will pave the way.

PR #970, opened 2024-06-20
move API to optional dependency group; move OpenData to default installation

rkingsbury:
WIP changes to dependency setup; checking impact on CI tests. See #928 .


rkingsbury:
Sorry all, I did not intend to merge this yet. (Perils of using GitHub on my phone!) I'll revert if @munrojm objects to these changes. It hasn't been released so shouldn't affect any users yet.

PR #1020, opened 2024-12-27
[Feature] Allow Different Azure Authentication Methods

jmmshn:
## Allow Different Azure Authentication Methods

Currently we only use `DefaultAzureCredentials` to auth.
There are some cases with custom auth rules where relying on the automatically resolving auth chain of `DefaultAzureCredentials` is too cumbersome and it would be nice to have the option to bypass it.
This PR allows us to directly import different Azure authentication mechansims and use them to authentication on Azure.

The changes here allows the user to just provide a string representation of the credential type for easy serialization, but does technically allow for arbitrary credential classes to be provided.

### DIFF
--- a/src/maggma/stores/azure.py
+++ b/src/maggma/stores/azure.py
@@ -41,10 +41,15 @@ CredentialType = Literal[
 ]
 
 
-def _get_azure_credential(credential_class: str):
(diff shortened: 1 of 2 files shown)


rkingsbury:
Thank you @jmmshn !

PR #1029, opened 2025-01-29
Remove python 3.8 classifier from pyproject.toml

Andrew-S-Rosen:
Only python 3.9+ is supported, so let's remove "Programming Language :: Python :: 3.8".


rkingsbury:
Thanks @Andrew-S-Rosen

PR #1033, opened 2025-02-26
fixed other references to incorrect method name

rekumar:
## Summary

I fixed the code examples in the quickstart docs

Major changes:

- fix: There was a typo in the quickstart docs for the Builder code that did not raise an error but also rendered the example incorrect. The example subclass of Builder implemented a method `.process_items` instead of the correct method `.process_item`, resulting in no processing being applied. Hopefully this can save others some headache when getting started :)

## Checklist

- [na] Google format doc strings added.
- [na] Code linted with `ruff`. (For guidance in fixing rule violates, see [rule list](https://beta.ruff.rs/docs/rules/))
- [na] Type annotations included. Check with `mypy`.
- [na] Tests added for new features/fixes.
- [ na] I have run the tests locally and they passed.

Tip: Install `pre-commit` hooks to auto-check types and linting before every commit:

```sh
pip install -U pre-commit
pre-commit install
```

### DIFF
--- a/docs/getting_started/simple_builder.md
+++ b/docs/getting_started/simple_builder.md
@@ -116,7 +116,7 @@ Our simple process item just has to multiply one field by `self.multiplier`:
 
 ``` python
 
-    def process_items(self, item : Dict) -> Dict:
+    def process_item(self, item : Dict) -> Dict:
         """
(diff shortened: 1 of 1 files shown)


rkingsbury:
Thanks for catching, and fixing, @rekumar !

PR #1037, opened 2025-03-09
Fix GridFSURIStore and increase testing

gpetretto:
While testing GridFSURIStore I discovered that there are several potential sources of errors due to the fact that it uses methods of the baseclass (GridFSStore), even if some attributes are not defined. Also there are no safeguard for accessing the `_files_store` attribute in case the store is not connected. 

This PR addresses these issues and modifies the gridfs tests so that all the tests are executed on both the `GridFSStore` and `GridFSURIStore`. This should hopefully reduce the chances of regression.

Also I noted that the `ssh_tunnel` is not taken into account in either of the "URI" stores. I kept the argument in the `__init__` to maintain the same API as the base classes, but it explicitly raises if passed.


rkingsbury:
Thanks @gpetretto

PR #1043, opened 2025-06-14
Remove Mongo Pin

CompRhys:
MP ecosystem is pinned to pymongo 11 trying to trace why and resolve.


rkingsbury:
This adds motivation to completing work on #846

PR #1111, opened 2026-06-21
Drop Python 3.9/3.10 support, add Python 3.13

rkingsbury:
- Updates requires-python to >=3.11
- Updates classifiers to reflect supported versions (3.11-3.14)
- Updates CI test matrix and upgrade-dependencies matrix to 3.11-3.14
- Updates docs badge and text to say Python 3.11+
- Removes requirements files for 3.9/3.10, adds placeholder requirements files for 3.13/3.14 (the upgrade-dependencies workflow will regenerate these with correct pinned versions)

NOTE: This PR was generated by Anthropic's Claude Code AI and reviewed by a human maintainer.


rkingsbury:
We observed that `python 3.14` kept hanging on the projection builder test, motivating some additional modernization work summarized below (by Claude):

**Root cause hypothesis:** Python 3.14's sys.monitoring API (used by coverage.py for instrumentation) may have a bug or edge case when a generator frame is suspended at a yield point while a local variable still holds a reference to another (exhausted) generator. Projection_Builder.get_items() was uniquely susceptible because it kept docs = store.query(...) as a named local variable at the yield point. Every other builder (MapBuilder, GroupBuilder) either calls list(store.query(...)) or inlines the query directly in the for loop.

Fixes made:

- [Primary fix] [projection_builder.py](vscode-webview://09k3vfqtg9m98j53h47n4vv9occ1d4c6taanqfnet4un29b262f6/src/maggma/builders/projection_builder.py): Changed docs = store.query(...); for d in docs: → for d in store.query(...): — eliminates the named generator variable from the frame at yield time, making it structurally identical to GroupBuilder's pattern.


- [Cleanup + correctness] All datetime.utcnow() calls replaced with datetime.now(UTC) across 7 source files and test_copy_builder.py. Ruff further simplified these to use from datetime import UTC (UP017 rule), the preferred 3.11+ idiom.
- [Diagnostic safety net] Added pytest-timeout==2.4.0 to testing deps and --timeout=120 to pytest addopts. If any test still hangs, it will now fail with a traceback showing exactly where it's stuck instead of keeping the CI job alive until the 7-minute job-level cancellation.

PR #1115, opened 2026-06-22
Remove pymongo<4.11 pin and switch to mongomock-ng

rkingsbury:
mongomock is unmaintained; [mongomock-ng](https://github.com/engFelipeMonteiro/mongomock-ng) is an actively maintained fork that supports newer pymongo releases, so the pymongo version ceiling can be dropped as well.

### Changes:
- replace `mongomock` with `mongomock-ng`
- remove `pymongo` version pin so that the latest releases can be used
- Bump the MongoDB server used in CI testing to be consistent with the minimum required pymongo version (4.2)


rkingsbury:
Fixes concerns raised in #1043

Issue #1113, opened 2026-06-22
[Bug]: Issue with `AzureBlobStore`  in maggma 0.73.0


Andrew-S-Rosen:
### Code snippet

```python

```

### What happened?

@rkingsbury, I happened to just update maggma the second you did, and I am getting the following error when importing jobflow. Should I open an issue on jobflow or is this a maggma-specific issue? Looks like a type hinting issue in maggma?

```python
______________ ERROR collecting tests/jobflow/test_copy_files.py _______________
tests/jobflow/test_copy_files.py:5: in <module>
    jf = pytest.importorskip("jobflow")
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/opt/hostedtoolcache/Python/3.12.13/x64/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
<frozen importlib._bootstrap>:1387: in _gcd_import
    ???
<frozen importlib._bootstrap>:1360: in _find_and_load
    ???
<frozen importlib._bootstrap>:1331: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:935: in _load_unlocked
    ???
<frozen importlib._bootstrap_external>:999: in exec_module
    ???
<frozen importlib._bootstrap>:488: in _call_with_frames_removed
    ???
/opt/hostedtoolcache/Python/3.12.13/x64/lib/python3.12/site-packages/jobflow/__init__.py:11: in <module>
    from jobflow.settings import JobflowSettings
/opt/hostedtoolcache/Python/3.12.13/x64/lib/python3.12/site-packages/jobflow/settings.py:7: in <module>
    from maggma.stores import MemoryStore
/opt/hostedtoolcache/Python/3.12.13/x64/lib/python3.12/site-packages/maggma/stores/__init__.py:6: in <module>
    from maggma.stores.azure import AzureBlobStore
/opt/hostedtoolcache/Python/3.12.13/x64/lib/python3.12/site-packages/maggma/stores/azure.py:56: in <module>
    class AzureBlobStore(Store):
/opt/hostedtoolcache/Python/3.12.13/x64/lib/python3.12/site-packages/maggma/stores/azure.py:392: in AzureBlobStore
    def _get_container(self) -> ContainerClient | None:
[...]

PR #1117, opened 2026-06-22
ci: add parallel test job for minimum (default) dependencies

rkingsbury:
## Summary

- Adds a new `test-min-deps` CI job that runs alongside the existing `test` job
- Installs only `pip install -e ".[testing]"` (base deps + test tools), omitting optional extras like `azure`, `vault`, `mongogrant`, and `vasp`
- Covers Python 3.11, 3.12, and 3.13 on ubuntu-latest
- Does not spin up the Azurite service (not needed without azure extras)
- Also wraps `mongogrant` imports in `test_advanced_stores.py` with a `try/except ImportError` + `pytest.skip(allow_module_level=True)` so the module skips cleanly when mongogrant is not installed (the `test_azure.py` file already handled the missing-azure case)

## Motivation

The existing `test` job verifies that all optional integrations work. This new job ensures the package is importable and core functionality works even when users install without any extras — catching regressions like #1113 earlier.

## Test plan

- [ ] `test-min-deps` job passes on all three Python versions
- [ ] `test` (full extras) job is unaffected

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rkingsbury:
With this PR, we will now be in a better position to catch bugs like #1113

